PR #24, opened 2020-12-22
feat: Add Doc states

kylecarbs:


cmoog (approved):
Something to consider: maybe it'd be best if we hid deprecated articles from the sidebar nav but still allowed direct links (from the product or other articles) to work properly.

kylecarbs:
100% agree. Neeeato

PR #26, opened 2020-12-23
feat: add admin docs for CVMs

cmoog:
This will replace the existing CVM article in the help center. We can follow-up this PR with cloud-specific deployment docs in the help center and link to them from here.

### DIFF
--- a/admin/environment-management/cvms.md
+++ b/admin/environment-management/cvms.md
@@ -10,7 +10,7 @@ programs inside their Environment including Docker and systemd.
 
 Navigate to **Manage > Admin > General** to enable this option.
 
-## Infrastructure requirements
+## Infrastructure Requirements
 
 - The Kubernetes Nodes must have a minimum kernel version of `5.4`
 (released Nov 24th, 2019).
--- a/admin/environment-management/cvms.md
+++ b/admin/environment-management/cvms.md
@@ -1,0 +1,86 @@
+---
+title: Docker in Environments
+description: Learn how to enable support for secure Docker inside Environments. 
+---
+
+As a platform administrator, you have the option to enable
+[Container-based Virtual Machines](../../environments/cvms.md) as an Environment
+deployment option. This option allows users to run system-level
+programs inside their Environment including Docker and systemd.
+
+Navigate to **Manage > Admin > General** to enable this option.
+
+## Infrastructure requirements
+
+- The Kubernetes Nodes must have a minimum kernel version of `5.4`
+(released Nov 24th, 2019).
+- The Kubernetes Nodes must be running an Ubunutu OS.
+- Legacy versions of cluster-wide proxy services such as Istio are not
+supported.
+- The cluster must allow privileged containers and `hostPath` mounts. Read more
+about why this is still secure [here](#security).
+
+## Security
+
+The [Container-based Virtual Machine](../../environments/cvms.md) deployment
+option leverages the [sysbox container runtime](https://github.com/nestybox/sysbox)
+to offer a VM-like user experience with the footprint of a typical container.
+
+Coder first launches a supervising container with privileged. This container is
+standard and included as part of the Coder release package. As part of the
+Environment build process, the supervising container launches an inner container
+using the [sysbox container runtime](https://github.com/nestybox/sysbox).
+This inner container is the user’s [Environment](../../environments/index.md).
+The user cannot gain access to the supervising container at any point during the
+lifecycle of their Environment. The isolation between the user’s Environment
+container and it’s outer, supervising container is what provides
+[strong isolation](https://github.com/nestybox/sysbox/blob/master/docs/user-guide/security.md).
+
+## Image Configuration
+
+### systemd
+
+During Environment startup, Coder checks for the presence of `/sbin/init` within
+the Environment Image. If the file exists, it's used as the container entrypoint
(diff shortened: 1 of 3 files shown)


admin/environment-management/cvms.md
```diff
@@ -0,0 +1,86 @@
...
+deployment option. This option allows users to run system-level
+programs inside their Environment including Docker and systemd.
+
+Navigate to **Manage > Admin > General** to enable this option.
+
+## Infrastructure requirements
+
+- The Kubernetes Nodes must have a minimum kernel version of `5.4`
+(released Nov 24th, 2019).
+- The Kubernetes Nodes must be running an Ubunutu OS.
```
nathanpotter:
```suggestion
- The Kubernetes Nodes must be running an Ubuntu OS.
```

nathanpotter: Is ubuntu the only supported option for sysbox?

admin/environment-management/cvms.md
```diff
@@ -0,0 +1,86 @@
...
+- Legacy versions of cluster-wide proxy services such as Istio are not
```
nathanpotter, line 18: Do you know which versions specifically?

admin/environment-management/cvms.md
```diff
@@ -0,0 +1,86 @@
...
+deployment option. This option allows users to run system-level
+programs inside their Environment including Docker and systemd.
+
+Navigate to **Manage > Admin > General** to enable this option.
+
+## Infrastructure requirements
+
+- The Kubernetes Nodes must have a minimum kernel version of `5.4`
+(released Nov 24th, 2019).
+- The Kubernetes Nodes must be running an Ubuntu OS.
```
nathanpotter, line 17: Is this a sysbox restriction or are we trying to limit the OSes we need to support?

admin/environment-management/cvms.md
```diff
@@ -0,0 +1,86 @@
...
+lifecycle of their Environment. The isolation between the user’s Environment
+container and it’s outer, supervising container is what provides
+[strong isolation](https://github.com/nestybox/sysbox/blob/master/docs/user-guide/security.md).
+
+## Image Configuration
+
+### systemd
+
+During Environment startup, Coder checks for the presence of `/sbin/init` within
+the Environment Image. If the file exists, it's used as the container entrypoint
```
nathanpotter, line 44: I think we should specify what happens if `/sbin/init` doesn't exist, which I believe is we just start our own init process – tinyinit or something.

admin/environment-management/cvms.md
```diff
@@ -0,0 +1,86 @@
...
+RUN ln -s /lib/systemd/systemd /sbin/init
+```
+
+### Adding Docker
+
+Be sure to install the `docker` packages into your Environment Image. For a
+seamless experience, use [systemd](#systemd) and register the `docker` service
+so `dockerd` is automatically run during initialization.
+
+The following snippet demonstrates how an image can registry the `docker`
```
nathanpotter, line 67:
```suggestion
The following snippet demonstrates how an image can register the `docker`
```

admin/environment-management/cvms.md
cmoog: @kylecarbs doesn't think this is *needed* but if I recall correctly the sysbox maintainers suggested it as a requirement. So including it in our docs feels like the safe bet either way.

kylecarbs: So with the `5.4` restriction I don't believe we need to limit to Ubuntu distributions... I can say though that this is all we've tested on, so might be risky to do anything else.

admin/environment-management/cvms.md
kylecarbs: Istio <= 1.14 is supported. @deansheather has better intel on these verisons.

cmoog: The phrasing here is intentionally vague until we have additional confidence for the combinations of Istio version and Kube version that work properly. Basically, not exactly

admin/environment-management/cvms.md
nathanpotter: agree, let's specify it and someone can ask for another distro if they want it – at that point we can test it out if it makes sense

admin/environment-management/cvms.md
cmoog: So technically we use `sleep infinity` as the entrypoint if no `/sbin/init` is found. This ensures that the Coder init is always executed as a container exec in both code paths...

cmoog: Either way, the exact behavior in the case of no `/sbin/init` feels like an implementation detail.

admin/environment-management/cvms.md
deansheather: Istio 1.5 or lower

Or newest Istio and k8s 1.19

deansheather: Actually I meant 1.4 or lower. My bad

deansheather: Actually it's 1.4 or lower

cmoog: Let's keep it vague until we're more confident from our own testing or until we have evidence from more customers that it's compatible with their cluster configurations.

kylecarbs (approved):

PR #28, opened 2021-01-04
Add missing namespace in kubectl command

sestegra:
If namespace `coder` is used to deploy coder, the namespace should be set in `kubectl log` command.


cmoog (approved):
Thank you for the contribution @sestegra!  🚀

PR #40, opened 2021-01-11
feat: add gcloud and eksctl example for kernel 5.4 nodes

cmoog:


admin/environment-management/cvms.md
```diff
@@ -21,6 +21,39 @@ supported.
...
+```yaml
+apiVersion: eksctl.io/v1alpha5
+kind: ClusterConfig
+metadata:
+  region: us-west-2
+  version: "1.17"
+  ...
+nodeGroups:
+  - name: coder-node-group
+    amiFamily: Ubuntu2004
```
deansheather, line 52: We haven't tested with 20.04, our internal config uses 18.04 (which is still supported).

admin/environment-management/cvms.md
```diff
@@ -21,6 +21,39 @@ supported.
...
+- GKE Master version `>= 1.17`
+- `node-version >= 1.17` and `image-type = "UBUNTU"`
+
+#### Example with `gcloud`
+
+```bash
+gcloud beta container node-pools create "coder-pool" --cluster <cluster> \
+    --image-type "UBUNTU" \
+    --node-version "1.17.14-gke.1600" \
+    ...
```
deansheather, line 35: We should probably give the cluster create command rather than the pool create command. There's an example in `setup_k8s.sh` in the monorepo.

admin/environment-management/cvms.md
cmoog: Ah ok cool. Not sure where that came from then. Will update.

admin/environment-management/cvms.md
```diff
@@ -21,6 +21,38 @@ supported.
 - The cluster must allow privileged containers and `hostPath` mounts. Read more
 about why this is still secure [here](#security).
 
+### Google Cloud Platform (GKE)
```
kylecarbs, line 24: Bit weird to abbreviate this with a different acronym...

Google Cloud Platform = GCP
Google Kubernetes Engine = GKE

admin/environment-management/cvms.md
```diff
@@ -21,6 +21,38 @@ supported.
...
+#### Example with `gcloud`
+
+```bash
+gcloud beta container clusters create "coder-cluster" \
+    --image-type "UBUNTU" \
+    --node-version "1.17.14-gke.1600" \
+    ...
+```
+
+### Amazon Web Services (EKS)
```
kylecarbs, line 38: Same comment as above here...

Maybe we do:
```
AWS: Elastic Kubernetes Service
GCP: Google Kubernetes Engine
```

admin/environment-management/cvms.md
cmoog: ok, maybe it should be `Google Cloud Platform w/ GKE`

admin/environment-management/cvms.md
```diff
@@ -21,6 +21,38 @@ Navigate to **Manage > Admin > Infrastructure** to enable this option.
...
+    --image-type "UBUNTU" \
+    --node-version "1.17.14-gke.1600" \
+    ...
+```
+
+### Amazon Web Services w/ EKS
+
+- Node group `amiFamily >= "Ubuntu1804"`
+
+#### Example with `eksctl`
```
kylecarbs, line 42: A bit weird to mention the CLI without showing any CLI usage.

cmoog: Meh. What about 
```
#### Example with an `eksctl` config spec
```

cmoog:
Side note: the styles for inline code blocks could use some improvement.

kylecarbs (approved):

PR #55, opened 2021-01-21
coder.com: Add documentation for Public API

sreya:


admin/api.md
```diff
@@ -0,0 +1,14 @@
+---
+title: "Public API"
+description: Learn more about Coder's API.
+state: alpha
+---
+
+In order to enable our users to integrate Coder into their automated workflows
+we've documented our HTTP API for public use. The API is still under active
+development and breaking changes can be expected as we finalize the endpoints.
+Stable API routes will be placed under a `/api/v1` root path.
```
khorne3, line 10: How about

> To help you integrate Coder into your automated workflows, we've documented our public API.
>
> Please note that the API is under active development; expect breaking changes as we finalize the endpoints. We will place stable API routes under the `/api/v1` path.

?

khorne3 (approved):

cmoog (approved):

admin/api.md
sreya: I like that, combing both of y'all's suggestion I think we change

```we've documented our public API```

to

```we've documented our API```

but otherwise I'll apply your comments

PR #62, opened 2021-01-26
fix: Cleanup 1.15.0, 1.15.1, 1.15.2

vapurrmaid:
- Normalize dates
- Normalize use of period (.) punctuation
- Add 1.15.2 to index


deansheather (approved):
We should add a template file and a bash script that templates it with the version and current date

vapurrmaid:
> We should add a template file and a bash script that templates it with the version and current date

Yep, just filed this earlier! #63

PR #64, opened 2021-01-28
chore: edit changes to docs

khorne3:
Docs affected:

* Compute Resources ~(also moved this out of admin since it's a concept doc)~
* SSH
* Editors and IDE


cmoog:
Let's get a process for adding redirects so we don't break these links

khorne3:
@cmoog I was told I could add entries to [this](https://github.com/cdr/m/blob/6f3e3490afd5201590aede068639b07be0798604/product/coder.com/vercel.json) for redirects; is that not the case or is there a better way?

cmoog:
Oh that works great, nice 👍

cmoog (approved):
Great changes 👍

PR #70, opened 2021-01-29
chore: port, reorg K8, install/setup docs

khorne3:
* Ported K8 docs from Zendesk and updated with better instructions
* Modified Install and Config guides so that the former gets you all the way through accessing Coder from the browser. Also added brief helm chart instructions.

### DIFF
--- a/setup/installation.md
+++ b/setup/installation.md
@@ -3,11 +3,12 @@ title: "Installation"
 description: Learn how to install Coder onto your infrastructure.
 ---
 
-This article walks you through the process of installing Coder.
+This article walks you through the process of installing Coder onto your
+[Kubernetes cluster](kubernetes/index.md).
 
 ## Dependencies
 
-Install the following dependencies:
+Install the following dependencies if you haven't already:
 
 - [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
 - [helm](https://helm.sh/docs/intro/install/)
@@ -15,6 +16,22 @@ Install the following dependencies:
 **For Production deployments:** set up and use an external
 [PostgreSQL](https://www.postgresql.org/docs/12/admin.html) instance to store
 data, including environment information and session tokens.
+
+## Creating the Coder Namespace (Optional)
+
+We recommend running Coder in a separate
+[namespace](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/);
+to do so, run
+
+```bash
+kubectl create namespace coder
+```
+
+Next, change the kubectl context to point to your newly created namespace:
+
+```bash
+kubectl config set-context --current --namespace=coder
+```
 
 ## Installing Coder
 
@@ -24,14 +41,28 @@ data, including environment information and session tokens.
    helm repo add coder https://helm.coder.com
    ```
 
-2. Install the helm chart into the cluster
+2. Install the helm chart onto your cluster (see the
+   [changelog](../changelog/index.md) for a list of Coder versions)
 
    ```bash
-   kubectl create namespace coder
-   helm install --namespace coder coder coder/coder --version <VERSION-NUMBER>
+   helm install --namespace coder coder coder/coder
    ```
 
-3. **For Production deployments:** Add the following to your helm chart so that
+3. Get a copy of your helm chart so that you can modify it; you'll need to
+   modify the helm chart to update your PostgreSQL databases (step 4) and enable
+   Dev URLs (step 5):
+
+   a. Get a copy of your existing helm chart and save as `values.yaml`: `helm
+   show values coder/coder > values.yaml`
+
+   b. Edit the `values.yaml` file as needed.
+
+   c. Upgrade/install your Coder deployment with the updated helm chart (be sure
+      to replace the placeholder value with your Coder version): `helm upgrade
+      coder coder/coder -n coder --version=<VERSION> -f values.yaml`. **This
+      must be done for whenever you update the helm chart.**
+
+4. **For Production deployments:** Add the following to your helm chart so that
    Coder uses your external PostgreSQL databases:
 
    ```yaml
@@ -48,11 +79,12 @@ data, including environment information and session tokens.
    You can find/define these values in your [PostgreSQL server configuration
    file](https://www.postgresql.org/docs/current/config-setting.html).
 
-4. [Enable Dev URL Usage](../admin/devurls.md). Dev URLs allow users to access
+5. [Enable Dev URL Usage](../admin/devurls.md). Dev URLs allow users to access
    the web servers running in your environment. To enable, provide a wildcard
-   domain and its DNS certificate and update your helm chart accordingly.
+   domain and its DNS certificate and update your helm chart accordingly. This
+   step is **optional** but recommended.
 
-5. After you've created the pod, tail the logs to find the randomly generated
+6. After you've created the pod, tail the logs to find the randomly generated
    password for the admin user
 
    ```bash
@@ -74,3 +106,23 @@ data, including environment information and session tokens.
 > If you lose your admin credentials, you can use the [admin password
 > reset](https://help.coder.com/hc/en-us/articles/360057772573) process to
 > regain access.
+
+## Accessing Coder
+
+1. To access Coder's web UI, you'll need to get its IP address by running the
+   following in the terminal to list the Kubernetes services running:
+
+   ```bash
+   kubectl --namespace coder get services
+   ```
+
+   The row for the **ingress-nginx** service includes an **EXTERNAL-IP** value;
+   this is the IP address you need.
+
+2. In your browser, navigate to the external IP of ingress-nginx.
+
+3. Use the admin credentials you obtained in this installation guide's previous
+   step to log in to the Coder platform. If this is the first time you've logged
+   in, Coder will prompt you to change your password.
+
+At this point, you're ready to proceed to [configuring Coder](configuration.md).
--- a/setup/installation.md
+++ b/setup/installation.md
@@ -55,7 +55,8 @@ kubectl config set-context --current --namespace=coder
    a. Get a copy of your existing helm chart and save as `values.yaml`: `helm
    show values coder/coder > values.yaml`
 
-   b. Edit the `values.yaml` file as needed.
+   b. Edit the `values.yaml` file as needed. Be sure to remove the lines that
+   you are *not* modifying, otherwise Coder will overwrite these values.
 
    c. Upgrade/install your Coder deployment with the updated helm chart (be sure
       to replace the placeholder value with your Coder version): `helm upgrade
--- a/setup/installation.md
+++ b/setup/installation.md
@@ -56,7 +56,8 @@ kubectl config set-context --current --namespace=coder
    show values coder/coder > values.yaml`
 
    b. Edit the `values.yaml` file as needed. Be sure to remove the lines that
-   you are *not* modifying, otherwise Coder will overwrite these values.
+   you are *not* modifying, otherwise the contents of `values.yaml` will
+   override those in the default chart.
 
    c. Upgrade/install your Coder deployment with the updated helm chart (be sure
       to replace the placeholder value with your Coder version): `helm upgrade
--- a/setup/kubernetes/aws.md
+++ b/setup/kubernetes/aws.md
@@ -1,0 +1,162 @@
+---
+title: Amazon Elastic Kubernetes Service
+description: Learn how to set up an Amazon EKS cluster for your Coder deployment.
+---
+
+This deployment guide shows you how to set up an Amazon Elastic Kubernetes
+Engine cluster on which Coder can deploy.
+
+## Prerequisites
+
+Please make sure that you have the following utilities installed on your
+machine:
+
+- [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
+- [AWS Command Line
+  Interface](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html)
+  (you'll also need to
+  [configure](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html)
+  the command-line interface to interact with your AWS account; consider AWS'
+  [CLI configuration
+  quickstart](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html)
+  to fast-track this process
+- [eksctl command line
+  utility](https://docs.aws.amazon.com/eks/latest/userguide/eksctl.html)
+
+## Preliminary Steps
+
+Before you can create a cluster, you'll need to perform the following to set up
+and configure your AWS account.
+
+1. Go to AWS' [EC2 Console](https://console.aws.amazon.com/ec2/); this should
+   take you to the EC2 page for the AWS region in which you're working (if not,
+   change to the correct region using the dropdown in the top-right of the page)
+2. In the **Resources** section in the middle of the page, click **Elastic
+   IPs**.
+3. Choose either an Elastic IP address you want to use or click **Allocate
+   Elastic IP address**. Choose **Amazon's pool of IPv4 addresses** and click
+   **Allocate**.
+4. Return to the EC2 Dashboard.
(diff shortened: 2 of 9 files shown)


setup/kubernetes/aws.md
```diff
@@ -0,0 +1,80 @@
...
+---
+
+This deployment guide shows you how to set up an Amazon Elastic Kubernetes
+Engine cluster on which Coder can deploy.
+
+## Prerequisites
+
+Before proceeding, please make sure that you have the [eksctl command line
+utility](https://docs.aws.amazon.com/eks/latest/userguide/getting-started-eksctl.html)
+installed on your machine.
```
IAmHughes, line 13:
```suggestion
Before proceeding, please make sure that you have the [eksctl command line
utility](https://docs.aws.amazon.com/eks/latest/userguide/getting-started-eksctl.html)
and the [aws command line utility](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html)
installed on your machine.
```

This is a good port, but we will also need to have the aws cli, I put a suggested change, but please feel free to modify it as needed.

IAmHughes (changes requested):
I left a suggested change for the AWS EKS install doc. Basically you also need `aws` cli installed and configured, so I linked to the install instructions from AWS.

Everything else looks great at the initial view. I haven't done a GKE or AKS deploy yet so I can't comment on those pieces

setup/kubernetes/aws.md
IAmHughes: My suggestion may be off on the 80 char line limit, webUI makes that a little hard sometimes.

setup/kubernetes/google.md
```diff
@@ -0,0 +1,104 @@
...
+CLI](https://cloud.google.com/sdk/docs/quickstarts) is installed on your
+machine.
+
+The following will spin up a Kubernetes cluster using the `gcloud` command (be
+sure to replace the parameters and environment variables as needed to reflect
+the needs of your environment):
+
+```bash
+PROJECT_ID="MY_PROJECT_ID" CLUSTER_NAME="MY_CLUSTER_NAME" \
+  gcloud beta container --project "$PROJECT_ID" \
```
kylecarbs, line 47: This command is a bit outdated. It doesn't support CVMs.

We should try to minimize the amount of flags too for customer sanity. What do you think about only documenting the CLI route? Customers can infer from the CLI options to adjust via the UI.

setup/kubernetes/azure.md
```diff
@@ -0,0 +1,79 @@
...
+
+```bash
+# You may have to run `az extension add --name aks-preview`
+#
+# You may also need to create a service principal manually using
+# `az ad sp create-for-rbac --skip-assignment`, then setting the
+# --service-principal and --client-secret flags
+
+CLUSTER_NAME="MY_CLUSTER_NAME" \
+LOCATION="##" SUBSCRIPTION="##" RESOURCE_GROUP="##" \
```
kylecarbs, line 39: I've never set Coder up on Azure before... but we should ensure this works with CVMs.

setup/kubernetes/index.md
```diff
@@ -0,0 +1,16 @@
...
+title: Kubernetes
+description: Learn how to set up a K8 cluster for your Coder deployment.
+---
+
+You can install and deploy Coder onto any Kubernetes cluster (version 1.13.7+)
+that meets the [requirements](../requirements.md).
+
+To help you get up and running, Coder offers the following deployment guides:
+
+- [Amazon Elastic Kubernetes Engine](../kubernetes/aws.md)
```
kylecarbs, line 11: Instead of having the bullet points here you can use:
```
<children></children>
```
to display all child topics.

kylecarbs (changes requested):

setup/kubernetes/aws.md
```diff
@@ -0,0 +1,82 @@
...
+  --ssh-public-key "$SSH_KEY_PATH" \
+  --managed
+```
+
+## Step 2: Adjust the K8 Storage Class
+
+Once you've created the cluster, adjust the default Kubernetes storage class to
+support immediate volume binding.
+
+First, delete the gp2 storage class: `kubectl delete sc gp2`
```
IAmHughes, line 45: We actually should ensure that `kubectl` is pointed to the right context first just in case `eksctl` didn't set the context automatically. Right now `eksctl` does that, but if it changes in the future this could delete the wrong storage class.

`kubectl config current-context` which should return with the cluster you created.

setup/installation.md
```diff
@@ -24,10 +25,10 @@ data, including environment information and session tokens.
    helm repo add coder https://helm.coder.com
    ```
 
-2. Install the helm chart into the cluster
+2. Install the helm chart onto your cluster (see the
+   [changelog](../changelog/index.md) for a list of Coder versions)
 
    ```bash
-   kubectl create namespace coder
    helm install --namespace coder coder coder/coder --version <VERSION-NUMBER>
```
IAmHughes, line 32:
```suggestion
   helm install --namespace coder coder coder/coder
```

IAmHughes (changes requested):
Left another suggestion re: the `helm install` command. Removing `--version` will default to latest which makes sense to me for the "happy path" A-Z startup guide?

setup/kubernetes/aws.md
```diff
@@ -0,0 +1,82 @@
...
+
+```bash
+CLUSTER_NAME="MY_CLUSTER_NAME" \
+  SSH_KEY_PATH="~/.ssh/my-public-key.pub" REGION="us-east-2" \
+  eksctl create cluster \
+  --name "$CLUSTER_NAME" \
+  --version 1.17 \
+  --region "$REGION" \
+  --nodegroup-name standard-workers \
+  --node-type t3.medium \
```
IAmHughes, line 31: We may need to bump this up for a better experience. End users hopefully can scope their own node sizes, but a `t3.medium` is very limited (2 CPUs and 4 GB of RAM)

setup/kubernetes/google.md
```diff
@@ -0,0 +1,104 @@
...
+    "projects/${PROJECT_ID}/regions/us-central1/subnetworks/default" \
+    --default-max-pods-per-node "110" \ 
+    --addons HorizontalPodAutoscaling,HttpLoadBalancing \ 
+    --enable-autoupgrade \ 
+    --enable-autorepair \ 
+    --enable-network-policy \ 
+    --enable-autoscaling 
+    --min-nodes "2" 
+    --max-nodes "8" \
+    --region "us-central1-c"
```
IAmHughes, line 72:
```suggestion
  clusters create "$CLUSTER_NAME" \
   --zone "us-central1-a" \
   --no-enable-basic-auth \
    --cluster-version "1.14.7-gke.14" \
    --machine-type "n1-standard-4" \
    --image-type "COS" \
    --disk-type "pd-standard" \
    --disk-size "100" \
    --metadata disable-legacy-endpoints=true \
    --scopes "https://www.googleapis.com/auth/cloud-platform" \
    --num-nodes "2" \
    --enable-stackdriver-kubernetes \
    --enable-ip-alias \
    --network "projects/${PROJECT_ID}/global/networks/default" \
    --subnetwork \
    "projects/${PROJECT_ID}/regions/us-central1/subnetworks/default" \
    --default-max-pods-per-node "110" \
    --addons HorizontalPodAutoscaling,HttpLoadBalancing \
    --enable-autoupgrade \
    --enable-autorepair \
    --enable-network-policy \
    --enable-autoscaling \
    --min-nodes "2" \
    --max-nodes "8" \
    --region "us-central1-c"
```

We were missing some backslashes and had some trailing whitespaces that would mess up the command.

setup/kubernetes/google.md
khorne3: I think documenting the CLI route only is fine

setup/kubernetes/aws.md
khorne3: @IAmHughes would be happy to bump this to large or xlarge -- what do you think of doing so now?

IAmHughes: I think at least `large` but more so a callout about sizing would be ideal.

setup/kubernetes/aws.md
```diff
@@ -0,0 +1,118 @@
...
+
+Before proceeding, please make sure that you have the following utilities
+installed on your machine:
+
+- [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
+- [AWS Command Line
+  Interface](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html)
+  (you'll also need to
+  [configure](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html)
+  the the command line interface to interact with your AWS account; consider
```
IAmHughes, line 19:
```suggestion
  the command line interface to interact with your AWS account; consider
```

setup/kubernetes/aws.md
```diff
@@ -0,0 +1,118 @@
...
+  AWS' [CLI configuration
+  quickstart](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html)
+  to fast-track this process))
```
IAmHughes, line 22:
```suggestion
  to fast-track this process
```

I may be missing something but I think these are extra `))` that aren't needed? @khorne3 let me know :)

setup/kubernetes/aws.md
```diff
@@ -0,0 +1,118 @@
...
+- [eksctl command line
+  utility](https://docs.aws.amazon.com/eks/latest/userguide/eksctl.html)
+
+## Preliminaries
+
+1. Go to AWS' [EC2 Console](https://us-east-2.console.aws.amazon.com/ec2/); this
```
IAmHughes, line 28:
```suggestion
1. Go to AWS' [EC2 Console](https://console.aws.amazon.com/ec2/); this
```

Updating URL as we say it should take you to your AWS region you're working in, but it was set to a specific region in the URL. The above will redirect to the current region (aka last region you were using when logged in)

setup/kubernetes/azure.md
IAmHughes: I believe @zmcnelis-coder said there is an issue with CVMs on Azure not being supported currently? Do we want to call that out? Something to do with the kernel not being updated yet?

setup/kubernetes/google.md
```diff
@@ -0,0 +1,89 @@
...
+   --default-max-pods-per-node "110" \
+   --addons HorizontalPodAutoscaling,HttpLoadBalancing \
+   --enable-autoupgrade \
+   --enable-autorepair \
+   --enable-network-policy \
+   --enable-autoscaling
+   --min-nodes "1"
+   --max-nodes "8" \
+   --region "us-central1-c"
+```
```
IAmHughes, line 53:
```suggestion
```bash
gcloud beta container --project "PROJECT_ID" \
clusters create "NEW_CLUSTER_NAME"
   --zone "us-central1-a" \
   --no-enable-basic-auth \
   --cluster-version "latest" \
   --machine-type "n1-standard-4" \
   --image-type "COS" \
   --disk-type "pd-standard" \
   --disk-size "50" \
   --metadata disable-legacy-endpoints=true \
   --scopes "https://www.googleapis.com/auth/cloud-platform" \
   --num-nodes "2" \
   --enable-stackdriver-kubernetes \
   --enable-ip-alias \
   --network "projects/PROJECT_ID/global/networks/default" \
   --subnetwork \
   "projects/PROJECT_ID/regions/us-central1/subnetworks/default" \
   --default-max-pods-per-node "110" \
   --addons HorizontalPodAutoscaling,HttpLoadBalancing \
   --enable-autoupgrade \
   --enable-autorepair \
   --enable-network-policy \
   --enable-autoscaling \
   --min-nodes "1" \
   --max-nodes "8"
```
```

@khorne3 did you get this command to work? I had to change a few things, such as missing backslashes and then I never was able to specify both a region and a zone, it was one or the other. Also the cluster version didn't work for me when I tried it as the current docs say, I changed it to `latest` instead.

**Review by IAmHughes — CHANGES_REQUESTED** [review]
Left some more suggestions and open questions.

File: setup/kubernetes/google.md [hunk 13]
```diff
@@ -0,0 +1,89 @@
...
+
+To create clusters capable of supporting use of the
```
khorne3, line 55: @IAmHughes here's the link, and then below that is the actual snippet so that people don't have to go to the link

IAmHughes: Got it, thanks!

setup/kubernetes/azure.md
khorne3: Azure [requires hacky fixes](https://codercom.slack.com/archives/C014R6H5FK8/p1610561051231200) ATM

setup/kubernetes/google.md
khorne3: I updated to fix the region/zone thing and cluster version and to remove trailing whitespaces

setup/kubernetes/google.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Set Up the GKE Cluster
+
+The following will spin up a Kubernetes cluster using the `gcloud` command (be
+sure to replace the parameters (specifically `PROJECT_ID` and
+`NEW_CLUSTER_NAME`) as needed to reflect the needs of your environment).
+
+```bash
+gcloud beta container --project "PROJECT_ID" \
+clusters create "NEW_CLUSTER_NAME"
```
IAmHughes, line 23:
```suggestion
clusters create "NEW_CLUSTER_NAME" \
```

This just needs a backslash at the end since the lines below are continuations of the same command

setup/kubernetes/google.md
```diff
@@ -0,0 +1,83 @@
...
+   --network "projects/PROJECT_ID/global/networks/default" \
+   --subnetwork \
+   "projects/PROJECT_ID/regions/us-central1/subnetworks/default" \
+   --default-max-pods-per-node "110" \
+   --addons HorizontalPodAutoscaling,HttpLoadBalancing \
+   --enable-autoupgrade \
+   --enable-autorepair \
+   --enable-network-policy \
+   --enable-autoscaling
+   --min-nodes "1"
```
IAmHughes, line 45:
```suggestion
   --enable-autoscaling \
   --min-nodes "1" \
```

These need trailing `\` as they have additional lines under them for the same command.

IAmHughes (changes requested):
Just a couple suggestions on this review iteration. I made them suggested commits so you should just be able to accept the commit suggestion if you'd like

setup/installation.md
```diff
@@ -24,14 +41,28 @@ data, including environment information and session tokens.
...
 
-3. **For Production deployments:** Add the following to your helm chart so that
+3. Get a copy of your helm chart so that you can modify it; you'll need to
+   modify the helm chart to update your PostgreSQL databases (step 4) and enable
+   Dev URLs (step 5):
+
+   a. Get a copy of your existing helm chart and save as `values.yaml`: `helm
+   show values coder/coder > values.yaml`
+
+   b. Edit the `values.yaml` file as needed.
```
bpmct, line 58: We should indicate that people should remove the lines from `values.yaml` that they are not modifying. Otherwise, I'm pretty sure an upgrade may fail due to an unnecessary override. For example, some lines contain the Coder version number. We don't want to override that when updating.
```yaml
# values.yaml:146-150

envproxy:
  # envproxy.replicas -- The number of replicas to run of the envproxy.
  replicas: 1
  # envproxy.image -- Injected during releases.
  image: docker.io/coderenvs/coder-service:1.15.2
```

bpmct: To clarify, I believe _this file_ would override the values from Coder's helm chart. So, if the values.yaml has `1.15.2` and the default chart is trying to upgrade to `1.16.0`, it will still pull in the `1.15.2` image.

khorne3: @bpmct Do you think this is clearer? `Be sure to remove the lines that you are *not* modifying, otherwise
the contents of `values.yaml` will override those in the default chart.`

bpmct: Yes totally, thanks 🙏

IAmHughes (approved):
LGTM! :shipit: 🚀

PR #73, opened 2021-02-01
Fix path to Compute Resources

khorne3:
Path to compute resources is incorrect, causing preview issues


IAmHughes (approved):
Seems to have fixed preview, thanks!

PR #74, opened 2021-02-02
feat: add docs on OpenSSH support

cmoog:
[ch6202]

### DIFF
--- a/admin/environment-management/ssh-access.md
+++ b/admin/environment-management/ssh-access.md
@@ -9,7 +9,7 @@ private key which provides access to all of their enviornments.
 As part of the standard Coder envionrment asset bundle, a lightweight SSH
 server is mounted into the enviornment agent and used as a backup when no server
 is found on port 22. This allows slimmer images to
-remain accessible by SSH without any image dependencie. The built-in server
+remain accessible by SSH without any image dependencies. The built-in server
 is limited, and does not properly implement more advaned usages of SSH like
 X11-forwarding and `sshd_config` specifications.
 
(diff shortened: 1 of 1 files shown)


admin/environment-management/ssh-access.md
```diff
@@ -1,10 +1,46 @@
...
+
+As part of the standard Coder envionrment asset bundle, a lightweight SSH
+server is mounted into the enviornment agent and used as a backup when no server
+is found on port 22. This allows slimmer images to
+remain accessible by SSH without any image dependencies. The built-in server
+is limited, and does not properly implement more advaned usages of SSH like
+X11-forwarding and `sshd_config` specifications.
+
+If SSH is the primary mode of access for your users, consider running a full
+OpenSSH server with `systemd` inside your image instead.
```
kylecarbs, line 17: Good idea to link to CVM docs here.

kylecarbs (approved):

admin/environment-management/ssh-access.md
```diff
@@ -1,17 +1,61 @@
...
+```
 
 ## Disable SSH Access
 
-If you would like to disable SSH access, either:
+If you would like to disable SSH access, you can either:
 
 - Run `helm install --set ssh.enable=false`
-- Add the following to your values override file and run `helm install -f values.yaml`
+- Add the following to your helm chart and run `helm install -f values.yaml`
```
khorne3, line 58: @cmoog sometimes we recommend `helm install ...` and sometimes we recommend `helm upgrade ...` with chart changes; which was is better?

khorne3 (approved):

admin/environment-management/ssh-access.md
cmoog: I suppose `helm upgrade --install` is the most flexible but I don't think it matters too much.

PR #81, opened 2021-02-09
[cd] Update & push submodule in cdr/m on pushes to master

bpmct:
This should trigger a build of coder.com/docs automatically instead of waiting for dependabot.


.github/workflows/deploy.yaml
```diff
@@ -20,14 +19,13 @@ jobs:
       - name: Checkout Docs
         run: git fetch && git checkout ${GITHUB_SHA}
```
bpmct, line 20: I wasn't sure if we wanted to do `git checkout ${GITHUB_SHA}` or `git checkout master`

.github/workflows/deploy.yaml
```diff
@@ -20,14 +19,13 @@ jobs:
...
-          GITHUB_REF: ${{ github.ref }}
-          GITHUB_REPOSITORY: ${{ github.repository }}
-          VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
-          VERCEL_ORG_ID: team_tGkWfhEGGelkkqUUm9nXq17r
-          VERCEL_PROJECT_ID: QmZRucMRh3GFk1817ZgXjRVuw5fhTspHPHKct3JNQDEPGd
+      - name: Commit files
+          run: |
+            git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
+            git config --local user.name "github-actions[bot]"
+            git commit -m "cdr/docs: bump product/coder.com/site/docs ${GITHUB_SHA}" -a
```
bpmct, line 26: We can definitely change these. I was following https://github.com/marketplace/actions/github-push

kylecarbs:
I'm relatively hesitant to make this workflow automated... the fact that pushing content here could cause our monorepo to fail a build scares me.

What's the reason behind this change? If we need docs to update faster we could manually trigger dependabot in an action so changes get merged faster.

cmoog:
> I'm relatively hesitant to make this workflow automated... the fact that pushing content here could cause our monorepo to fail a build scares me.

So we could always revert the commit in the mono repo.... the integrity of version control in the mono repo is maintained with this change.

cmoog:
> What's the reason behind this change? If we need docs to update faster we could manually trigger dependabot in an action so changes get merged faster.

It would be cleaner for contributors if they didn't have to think about the mono repo at all.

kylecarbs:
Agreed agreed

bpmct:
Yeah, I don't like the idea that something like a typo fix could take up to 24h (+ approve time) to update on the site. Ideally, the docs should be automatically built after a push. 

We could potentially do this without pushing code to master on the monorepo, but then we probably shouldn't be importing our docs into the monorepo as a submodule anyways.

cmoog (approved):
The approach looks great to me 👍 . I assume this is something we can only _really_ test by merging.

kylecarbs (approved):

PR #92, opened 2021-02-12
Fix deploy-preview to allow `/` in branch name

IAmHughes:
This fixes the match criteria for branches to allow a `/` in the branch name, i.e. `iamhughes/test`

[See filter pattern cheat sheet](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet) for more info.


IAmHughes:
Actually, looking at this again I don't think this will fix the issue. I think `**` was correct. Closing w/o merging.

PR #97, opened 2021-02-16
add info about trying coder

bpmct:
- As someone evaluating Coder, I should know that Coder has a 60 day trial license that can be immediately generated on https://coder.com/trial
- As someone evaluating Coder, I should know that Coder has a "local preview" option that is documented [here](https://codercom-km0xqlzsl.vercel.app/docs/setup/kubernetes/local-preview) (#68) and does not require a license.  

I am thinking that this info should probably be root level with some text in the [Setup](https://coder.com/docs/setup) page so the user doesn't have to do a lot of clicking around. Let me know if you think another hierarchy would be better.

I like how Okteto does this: https://okteto.com/docs/enterprise/install/overview.

⚠️ Do not merge until #68 is merged.

PR #94, opened 2021-02-16
add guide for writing Coder images

bpmct:
This is still a draft, but I wanted to get started on a guide that covers how to add common elements (such as an IDE, VS Code extension, or git repository) to a Coder image.

I also wanted to mention Coder's example images repo, as well as how sysbox is necessary to build images for CVMs. This may be better suited in the "Guides" section, and it definitely needs some editing. Any thoughts @khorne3?

PR #98, opened 2021-02-16
attempt to fix deploy previews on forks

bpmct:
my latest two prs (#97 and #94) did not run the action. I wasn't sure if it was intentional or not to disable previews from forks.

ref: https://futurestud.io/tutorials/github-actions-run-on-pull-request

We could potentially limit to PRs against master.

### DIFF
--- a/.github/workflows/deploy-preview.yaml
+++ b/.github/workflows/deploy-preview.yaml
@@ -4,6 +4,9 @@ on:
     branches:
       - "**"
       - "!master"
+    pull_request:
+      branches:
+        - "**"
 jobs:
   preview:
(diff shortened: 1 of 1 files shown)


cmoog (approved):
preview deploys also seem to be broken here https://github.com/cdr/docs/pull/78 which suggests there may be something else wrong... 🤷

bpmct:
that is potentially due to this other bug that @IAmHughes discovered:
> well that resolved it....wow...so renaming an action when you have a branch open will screw that branch up for all actions to be executed that changed name. thats really not good lol. i'll submit a bug to GitHub

PR #99, opened 2021-02-16
add trial info (test deploy?)

bpmct:


bpmct:
unsure how to fix deploys from forks :(

PR #105, opened 2021-02-17
chore: Update Image, Setup, Admin assets

khorne3:
Changes to docs due to UI updates. Also contains minor copyedits.

Do not merge before 1.16 release.


admin/appearance.md
```diff
@@ -12,27 +12,25 @@ Coder offers you appearance customization options for the following:
 
 To customize your system banner messages:
 
-1. Go to **Admin > Appearance** in the Coder UI.
+1. Go to **Manage** > **Admin** > **Appearance** in the Coder UI.
```
jawnsy, line 15: A little Markdown tip: if you use the 1. for all the bullets, then they will be autonumbered, which is helpful if you rearrange items. For example the following list uses 1. 1. 1.

1. List
1. Item
1. Other

admin/environment-management/index.md
```diff
@@ -5,18 +5,19 @@ description: Learn how to manage Environments from an admin level.
...
-- You can also modify the [Environment Shutdown
-  Behavior](shutdown.md) on a per-organization basis to optimize
-  resource usage.
+- You can also modify the [Environment Shutdown Behavior](shutdown.md) on a
+  per-organization basis to optimize resource usage.
 
-- You can [disable SSH access](ssh-access.md) for your
-  end-users by editing the YAML file in helm (by default, Coder enables SSH
-  access to environments).
+- You can [disable SSH access](ssh-access.md) for your end-users by editing the
```
jawnsy, line 20: your end users -> for users

YAML file in helm -> Helm values file

in the parenthetical: Coder permits SSH access to environments by default

admin/extensions.md
```diff
@@ -6,45 +6,27 @@ description: Learn how to configure an internal marketplace for air-gapped deplo
...
-### Adding an Extension to the Internal Marketplace
+If you run Coder in an air-gapped environment, the public VS Code marketplace is
+inaccessible to end-users. Using the **Custom** configuration option, you can
+point Coder to an air-gapped instance of
+[OpenVSX](https://github.com/eclipse/openvsx) to serve assets to users.
 
-You can add extensions to the internal marketplace by uploading the **.vsix**
-bundle. To do so, navigate to **Manage** > **Extensions** and select **Upload
-Extensions**. Find and upload the file for your extension.
+View the [OpenVSX deployment
```
jawnsy, line 30: Markdown tip: you can separate the link text from the link itself, which makes it easier to read the markup. For example, the same text can be produced with the following markup:

```markdown
View the [OpenVSX deployment wiki] for more information.

[OpenVSX deployment wiki]: https://github.com/eclipse/openvsx/wiki/Deploying-Open-VSX
```

This way the text will be easier for you to read, while preserving the link for people looking at the rendered output.

images/deprecating.md
```diff
@@ -5,13 +5,13 @@ description: Learn how to deprecate an image.
 
 You can mark images as deprecated to prevent them from being used to create new environments.
 
-1. Go to "Images" and find the image you want to deprecate.
-2. Click "Edit".
-3. Select the "Deprecate this image" checkbox.
-4. Click "Update Image" to save your changes.
+1. Go to **Images** and find the image you want to deprecate.
```
jawnsy, line 8: I'm not sure if this is better or worse, but: want to deprecate -> mark as deprecated?

images/index.md
```diff
@@ -1,11 +1,11 @@
...
-description: Learn about the base of reproducibility inside Coder.
+description: Learn about the basis of reproducibility inside Coder.
```
jawnsy, line 4: Suggest: Learn about reproducibility in Coder

images/index.md
```diff
@@ -1,11 +1,11 @@
...
 ---
 
 Images are the basis from which new [Environments](../environments/index.md)
-are created. They are the base of Environment reproducibility.
+are created. They are the basis of environment reproducibility.
```
jawnsy, line 8: Coder creates [Environments] using templates called _images_, which encourage reproducibility of development environments.

If we haven't already, we should blog about the benefits of reproducible software development practices, as I think that's a key differentiator for us as compared to VDI solutions like Citrix

images/tags.md
```diff
@@ -32,7 +34,7 @@ that image.
 
 ![Set Default Tag](../assets/default-tag.png)
 
-To use an existing tag as the default tag, click the 3 dot menu for a tag and
-select **Make default**.
+To use an existing tag as the default tag, click the three-dot menu for a tag
```
jawnsy, line 37: The downside of these kitchen-sink overflow menu buttons is that they are not super easy to describe. These are sometimes called the Pancake or [Hamburger button](https://en.wikipedia.org/wiki/Hamburger_button). I guess in this case it's an ellipsis button?

image

https://twitter.com/lukew/status/591296890030915585

khorne3: I think I've used vertical ellipses elsewhere, so I think I'll do so here? I think kebab menu might be confusing (or am I the only person who hasn't heard it called this?)

jawnsy: Oh, I think Luke was being a bit tongue-in-cheek with the names. Yeah vertical ellipsis makes sense. I was just making a general remark about these buttons being difficult to describe because they're a relatively recent (on the human scale of things) invention.

cmoog (approved):

PR #106, opened 2021-02-18
chore: port docs from Zendesk

khorne3:


guides/ssl-certificates.md
```diff
@@ -0,0 +1,216 @@
...
+spec:
+  secretName: coder-devurls-cert
+  duration: 2160h # 90d
+  renewBefore: 360h # 15d
+  dnsNames:
+     - "*.coder.example.com" # Your dev URLs wildcard subdomain
+issuerRef:
+  name: letsencrypt
+  kind: Issuer
+```
```
bpmct, line 188: This step isn't necessary and the certificates can be created automatically via the ingress resource. (ref: https://cert-manager.io/docs/usage/ingress/). 

To do this in Coder, in the step 4 yaml, use:

```yaml
ingress:
  additionalAnnotations:
  - 'cert-manager.io/cluster-issuer: nameofclusterissuer'
```

Following this example, the name of the issuer/cluster issuer is "letsencrypt." I have only tested this with a Cluster Issuer, but according to the cert-manager docs `cert-manager.io/issuer` can be used for an Issuer resource, which this tutorial uses instead.

cert-manager's ingress guide (linked above) as well as this separate doc have been SO helpful for me troubleshooting this stuff: https://cert-manager.io/docs/faq/troubleshooting/

cc @mterhar to include anything I may have missed.

environments/personalization.md
```diff
@@ -40,35 +42,67 @@ sudo apt-get install -y fish
...
+
+# Backup existing gitconfig if it exists
+if [ -f ~/.gitconfig ]; then
+echo "Backing up ~/.gitconfig"
+mv ~/.gitconfig ~/.gitconfig.bak
+fi
+
+# Set name and email in git
+echo "[user]\n\temail = youremailhere@gmail.com\n\tname = Your Name" > ~/.gitconfig
+```
```
bpmct, line 66: I wonder if stuff like this should be eventually be hosted in github gists or in a source-controlled "examples" repo and somehow embedded into the doc? Then, it could maybe be "expanded" too for cleanliness.

Not saying, we should do this now, but would love to hear your thoughts.

setup/nightly-releases.md
```diff
@@ -0,0 +1,115 @@
+---
+title: nightly-releases
+description: Learn how to obtain Coder's nightly releases.
+---
```
bpmct, line 4: Thoughts on this being a guide?

"switching to nightly releases"

bpmct (changes requested):
I'm super pumped for these docs to be more visible. Left one comment on the SSL/TLS one, since Mike has been helping me with that recently.

setup/nightly-releases.md
bpmct: reasoning is just trying to keep the sidebar as clean as possible

cmoog:
I changed the base branch to the 1.16 release branch. This will make cutting the release much easier.

environments/personalization.md
khorne3: Definitely think that's something to consider in the future; I think have examples of how we do things would be helpful for our users, especially when it comes to gotchas and the like -- internal tricks and stuff we've learned.

guides/ssl-certificates.md
khorne3: @bpmct I think I got this section correct, but could you please take a second look? Thanks!

admin/access-control/password-reset.md
```diff
@@ -0,0 +1,41 @@
...
+description: Learn how to reset Coder user passwords.
+---
+
+To reset a user's password:
+
+1. Go to **Manage** > **Users**.
+2. Find the user whose password you want to reset and click the vertical
+   ellipses to the right.
+3. Click **Reset password**. Coder will display a temporary password that you
+   can provide to the user.
```
vapurrmaid, line 12: There's technically a confirmation dialog now (didn't used to be). They're shown the temporary password with cancel/reset and only if they click reset on the dialog is the action dispatched

Screenshot from 2021-02-22 09-51-05

guides/ssl-certificates.md
bpmct: @khorne3 So your tutorial is correct and works, but I am pretty sure it isn't the recommended way to add Certificates.

Instead, they can be automatically created via the ingress resource when those three lines are added to the helm chart file.

guides/ssl-certificates.md
```diff
@@ -0,0 +1,177 @@
...
+```yaml
+ingress:
+  useDefault: true
+  host: "coder.example.com"
+  tls:
+    enable: true
+    hostSecretName: coder-root-cert
+    devurlsHostSecretName: coder-devurls-cert
+  additionalAnnotations:
+  - 'cert-manager.io/cluster-issuer: letsencrypt'
```
bpmct, line 171: this would be
```
  - 'cert-manager.io/issuer: letsencrypt'
```

since this tutorial covers using an "Issuer" resource.  Under the `### ClusterIssuers` section, we need to mention the user should switch to
```yaml
  - 'cert-manager.io/cluster-issuer: letsencrypt'
```

My bad, I should have made that clear.

bpmct (approved):
The SSL/TLS certificate section looks good to me 👍🏼

vapurrmaid (approved):

PR #108, opened 2021-02-22
clarify personalize commands

ericpaulsen:
hey @khorne3 - i noticed that the ~/personalize script will abort if there is no -y flag when a command is run. i added the following lines to clarify this for users. let me know what you think.


khorne3:
@ericpaulsen I think this is helpful; the only thing is I would move it up to line 43 (just below the sample script)

ericpaulsen:
done. thanks @khorne3

PR #109, opened 2021-02-22
Fix: Remove `pull_request` event from preview deploys

IAmHughes:
`pull_request` was added as an event type to trigger deploys to preview from forks to assist with the OSS contribution process. Unfortunately, we depend on a secret to access a private repository which is NOT passed to the fork for security purposes (GitHub security enforces this). Therefore, I'm removing the `pull_request` event entirely which should fix the issues we were seeing.

Note: Pushes to branches in PRs will still trigger builds

### DIFF
--- a/.github/workflows/preview.yaml
+++ b/.github/workflows/preview.yaml
@@ -20,8 +20,13 @@ jobs:
           token: ${{ secrets.GH_ACCESS_TOKEN }}
           submodules: true
           fetch-depth: 0
-      - name: Checkout Docs
+      - name: Checkout Docs - Push
+        if: ${{ github.event_name == 'push'}}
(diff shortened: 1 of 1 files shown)


bpmct (approved):
I was struggling with this last week. If this fixes it, I will be so happy. Thanks Thomas! 🙌🏼

bpmct:
Screen Shot 2021-02-22 at 11 05 43 AM

@IAmHughes it looks like the action is firing now, but the checkout syntax doesn't work for PRs as it does for pushes.

IAmHughes:
@bpmct as we discussed, when you give a 👍 I'll merge.

bpmct (approved):

PR #122, opened 2021-02-24
fix: command formatting

ericpaulsen:
small change to add code formatting to the helm command.

### DIFF
--- a/guides/teardown.md
+++ b/guides/teardown.md
@@ -8,7 +8,7 @@ This guide shows you how to tear down Coder and the cluster it is deployed on.
 > These instructions help you remove infrastructure created when following our
 [Kubernetes setup tutorials](../setup/kubernetes/index.md). They do not include
 teardown steps for any additional resources that you create.  If you need to keep
-your cluster, you can run helm uninstall coder, which deletes all Coder
+your cluster, you can run `helm uninstall coder`, which deletes all Coder
 services but retains environments and their associated disk space.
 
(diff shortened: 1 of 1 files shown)


vapurrmaid (commented):
I know this may seem like overkill to some, but I have a preference towards using code blocks for almost all code cases like this, as it the way the structure flows is visually appealing and really draws attention to the comand.

> you can run
>
> ```bash
> helm uninstall coder
> ```
> ....lorem ipsum and so on


See Above ^

ericpaulsen:
> I know this may seem like overkill to some, but I have a preference towards using code blocks for almost all of these cases. The structural flow is visually appealing and draws attention to the command.
> 
> > you can run
> > ```shell
> > helm uninstall coder
> > ```
> > 
> > 
> > ....lorem ipsum and so on
> 
> See Above ^

I see what you're saying, although it might look a bit funky when creating a code block in an indented text section. I'll submit a second commit to see what it looks like in preview.

ericpaulsen:
Thoughts?

vapurrmaid:
> Thoughts?

You're right, it didn't render as expected/as nicely. I think let's just go with what you originally had - sorry for throwing for a loop!

ericpaulsen:
> > Thoughts?
> 
> You're right, it didn't render as expected/as nicely. I think let's just go with what you originally had - sorry for throwing for a loop!

No worries! Appreciate the diligence.

khorne3 (approved):

PR #123, opened 2021-02-24
ericpaulsen/sysboxcompatibility

ericpaulsen:
added link to the `sysbox` distro compatibility.


ericpaulsen:
accidentally added the `fix: command formatting` commit to this branch. please ignore.

khorne3:
We have both a Ubuntu and kernel requirement starting in line 13, so wdyt about updating that to something like `The Kubernetes Nodes must be running Ubuntu with Kernel **5.4** or later` to make it clearer?

ericpaulsen:
> We have both a Ubuntu and kernel requirement starting in line 13, so wdyt about updating that to something like `The Kubernetes Nodes must be running Ubuntu with Kernel **5.4** or later` to make it clearer?

well the nodes don't necessarily need to be running Ubuntu, as `sysbox` supports other Linux distros such as Debian, Fedora, & CentOS. but I do think we can include the `sysbox` kernel compatibility in the `Infrastructure Requirements` section. what do you think?

khorne3:
> nodes don't necessarily need to be running Ubuntu

The way we have CVMs documented says they do (I'm assuming in spite of what sysbox requires) -- is this the case @cmoog? I think this will inform how we write up the reqs

cmoog (approved):
This seems more appropriate under the “Infrastructure Requirements” section

cmoog:
> The way we have CVMs documented says they do (I'm assuming in spite of what sysbox requires) -- is this the case @cmoog? I think this will inform how we write up the reqs

I may have been mistaken when I originally said it had to be Ubuntu strictly when the more precise statement would probably have been "X Ubuntu node type on GKE/EKS has the correct kernel". I think we were also playing it safe for the initial alpha release. The sysbox docs are the source of truth here in my view, so it's good thinking by @ericpaulsen to link to that. 

@kylecarbs may have views on this as well...

khorne3:
@ericpaulsen wdyt about the updated suggestion (I moved the requirement + link to the requirements section)

ericpaulsen:
I like this change... but we're now referencing sysbox before the user gets to the `Security` section where the sysbox context/explanation lives. I'm thinking we could move the `Security` section up to the top of the page to give the user context on the tech behind CVMs, before diving into the Requirements, Setting Up, etc. Thoughts?

khorne3:
@ericpaulsen That's true -- I think security should stay where it is just to make the reading experience a bit more fluid. Maybe the requirement is

> CVMs leverage the sysbox container runtime, so the Kubernetes Node must run a supported Linux distro with the minimum Kernel version (see Sysbox Distro Compatibility for more information)

?

ericpaulsen:
Love it.

khorne3 (approved):

PR #126, opened 2021-02-25
[WIP] WAC mvp initial doc

Emyrk:
Very basic intro to WAC to give some info on what's going on with WAC.


vapurrmaid (changes requested):
I think we should create a 1.17 branch for these items

Emyrk:
> I think we should create a 1.17 branch for these items

To clarify, this feature and docs are for 1.16

vapurrmaid:
> > I think we should create a 1.17 branch for these items
> 
> To clarify, this feature and docs are for 1.16

We didn't announce/highlight or add anything to the [changelog](https://coder.com/docs/changelog/1.16.0) regarding WAC for 1.16. I don't think sneaking it into 1.16 is the best approach when we're working towards a GA MVP for 1.17.

Emyrk:
> > > I think we should create a 1.17 branch for these items
> > 
> > 
> > To clarify, this feature and docs are for 1.16
> 
> We didn't announce/highlight or add anything to the [changelog](https://coder.com/docs/changelog/1.16.0) regarding WAC for 1.16. I don't think sneaking it into 1.16 is the best approach when we're working towards a GA MVP for 1.17.

@cmoog Can you rebuttal this?

cmoog:
I think throwing this on zendesk is the best approach. That'll give us _something_ to share with those waiting on the full feature rather than leaving them in the dark for another month.

khorne3:
We're trying to move off of Zendesk, so we're kind of in a bind. I think if we had https://github.com/cdr/m/pull/7159 then maybe we could publish to this repo but hide it?

khorne3:
@Emyrk Should we move this content into the newer docs you're working on?

Emyrk:
@khorne3 yea, these docs here are the `cli` docs, the docs I was working on was a more substantial release.

The new docs will supersede this one.

PR #131, opened 2021-03-01
Updated to reflect new Coder trial program

CoderErik:
Removed material that applied to the previous trial program.

### DIFF
--- a/setup/requirements.md
+++ b/setup/requirements.md
@@ -49,12 +49,12 @@ The use of Coder deployments requires a license that's emailed to you.
 
 ### Restrictions
 
-Deployments using the free level of Coder (1-10 users):
+Deployments using the free 60-day free trial of Coder:
 
 - **Must** be able to reach and use an outbound internet connection (at minimum,
   your deployment must be able to access **licensor.coder.com**)
-- Cannot be used to set up an air-gapped deployment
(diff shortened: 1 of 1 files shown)


bpmct (approved):

setup/requirements.md
```diff
@@ -49,12 +49,12 @@ The use of Coder deployments requires a license that's emailed to you.
 
 ### Restrictions
 
-Deployments using the free level of Coder (1-10 users):
+Deployments using the free 60-day free trial of Coder:
```
vapurrmaid, line 52: **question:** Is it necessary to quantify here? I imagine if we drop the quantifier, it's less prone to becoming outdated without sacrificing clarity. Not sure though, thought I'd just ask

vapurrmaid, line 52: `free 60-day free trial of Coder`

free is mentioned twice in an awkward manner

vapurrmaid (changes requested):

setup/requirements.md
khorne3: so `deployments using the free trial of Coder`?

CoderErik: Good point. Your phrasing works better. Main thing is to get rid of the 1-10 users language since that is no longer the case.

khorne3 (approved):

PR #132, opened 2021-03-01
chore: add clarifying notes to AWS EKS doc

khorne3:


deansheather (commented):
LGTM. While we're in AWS land we should add instructions for installing Calico which is required for network policy enforcement.

cmoog (approved):
Look good. +1 the point about Calico

khorne3:
@deansheather Added Calico; thanks!

PR #133, opened 2021-03-01
Modify changelog index

khorne3:
Screen Shot 2021-03-01 at 3 25 43 PM


khorne3:
@vapurrmaid I noticed today that our index and our sidebar were out of date, so wdyt about just using `<children>` so we don't have to manually update?

vapurrmaid:
> @vapurrmaid I noticed today that our index and our sidebar were out of date, so wdyt about just using `<children>` so we don't have to manually update?

LGTM for now, though I don't think it's infinitely scalable.

I'd hope we could bucket these off by `major` once `2.0` hits.

So the sidebar would be:

changelog
  - 1.x.x
    - 1.17.0
    - 1.16.1
    - so on
  - 2.x.x
  
  We can cross that bridge once we get there. At the current release pace we're ok right now

khorne3:
Hmm, yeah, I see your point. We could also archive the previous major releases so that we only see the current iteration, so to speak (so when 2.x and future releases arrive, we put 1.x into an archive folder)

vapurrmaid:
> Hmm, yeah, I see your point. We could also archive the previous major releases so that we only see the current iteration, so to speak (so when 2.x and future releases arrive, we put 1.x into an archive folder)

Also sounds viable :+1:

PR #134, opened 2021-03-01
keycloak

ericpaulsen:
Keycloak setup & config translated from [Nathan's google doc.](https://docs.google.com/document/d/15fvmD2pu8sc3ka1-RG9ap7RS57gmlgsAyrAJkTS6WhU/edit#) I've included all of the screenshots for now (there are quite a few), but not sure if it's best to keep them, due to the likelihood of Keycloak & SAML UI updates.

Open to feedback & clean up.

[Clubhouse story](https://app.clubhouse.io/coder/story/4826/add-doc-saml-oidc-via-keycloak)


khorne3 (approved):

IAmHughes:
The images in here, were they added to `git-lfs` or are they just in the repo? /cc @ericpaulsen

khorne3:
@IAmHughes Images live just in the repo

IAmHughes:
> @IAmHughes Images live just in the repo

Any reason not to use lfs? Git doesn't do well with binary files over time. [Docs](https://git-lfs.github.com/).

Probably not a HUGE deal, but every time you change an image it doesn't version control like plaintext since it's a binary file, and therefore both copies are saved in storage at full size. Over time, this causes repo bloat which can cause all sorts of issues between CI time to local workspace stuff.

I think we'd be fine, but just curious why we weren't using it 😋

khorne3:
@IAmHughes No reason, but that's a good idea to do so. I'll get this ball rolling

IAmHughes:
@khorne3 let me know if you need help!

PR #138, opened 2021-03-03
chore: add doc on managing .gitconfig

ericpaulsen:
initial draft for clarifying git config management in Coder. context - [ch-2590](https://app.clubhouse.io/coder/story/2590/how-can-i-manage-git-configuration?vc_group_by=day)


guides/gitconfig.md
```diff
@@ -0,0 +1,28 @@
+---
+title: Managing Git Configuration
+description: Learn how to manage your Git configuration in Coder.
```
vapurrmaid, line 3: I Think we can opt for brevity here, something along the lines of:

```
Learn how to configure Git with Coder
```

Though I'd ultimately leave that decision to @khorne3

guides/gitconfig.md
```diff
@@ -0,0 +1,28 @@
...
+---
+
+This guide will show you how to manage Git configuration through the system,
+global, and repo levels.
+
+## How to use the System and Global configurations
+
+For customers seeking organizational (system-level) `.gitconfig` defaults, we recommend
```
vapurrmaid, line 11: I'd suggest leaving out "For customers". Though again, defer to @khorne3 on final copy

guides/gitconfig.md
```diff
@@ -0,0 +1,28 @@
...
+adding a default configuration into the `/etc/gitconfig` file of each image associated
+to the organization:
```
vapurrmaid, line 13: I think it may be useful to give context for System configurations. Furthermore, we don't offer any concrete advice on how to set the file.

Here's a few suggestions:

1. Start with a context of git configurations that apply to an organization

```
For scenarios whereby a set of git configurations should apply across an organization,
we recommend using a System-level git configuration. System-level git configurations
are located at `/etc/gitconfig`. These settings are applied to each git repository, but may
be overriden using global or worktree git configurations.
```

2. There are various ways to apply `/etc/gitconfig`. We suggest `coder/configure` for
applying these settings to a development image. Otherwise, they may be applied
using `~/personalize` or some other mechanism such as an install script or file sync (we don't need to document such).

guides/gitconfig.md
```diff
@@ -0,0 +1,28 @@
...
+
+```bash
+[url "git@github.com:"]
+    insteadOf = https://github.com/
+```
```
vapurrmaid, line 18: This is not a `bash` code fence. This is `gitconfig`. Furthermore, the example should prefaced that this is an example `.gitconfig` file and the example should explain what it's doing.

guides/gitconfig.md
```diff
@@ -0,0 +1,28 @@
...
+
+For users seeking to override the system `.gitconfig` with a global setting of their
+own, we recommend including `.gitconfig` and other personal configuration files in
+the `home/coder` directory. [Learn more about environment personalization.](https://coder.com/docs/environments/personalization#dotfiles-repo)
```
vapurrmaid, line 22: We recommend setting a `.gitconfig` in personalize regardless. The fact that it overrides system settings is more pertinent when explaining how to apply gitconfig settings across an organization.

Thus the wording here should start with "We recommend ....". We can re-mention that it overrides system `etc/gitconfig` as an aside, but it's not the leading concept.

guides/gitconfig.md
ericpaulsen: thanks for the feedback @vapurrmaid - I can certainly add the context you provided above. I think that will be helpful in framing the guide for the user. 

Question - where can one edit `coder/configure` to apply System-level git defaults?

guides/gitconfig.md
ericpaulsen: This makes sense - so I'll clarify with something like:

> While we recommend setting a `.gitconfig` file via the `~/personalize` script, it will override the system-level `etc/gitconfig` default set in the image.

guides/gitconfig.md
IAmHughes:
```suggestion
Users seeking organizational (system-level) `.gitconfig` defaults, we recommend
```

agreed, would do something like this perhaps

guides/gitconfig.md
IAmHughes: Do we want to provide examples? There are a lot of guides out there on `gitconfig` and while we can provide one too, it's not necessarily our area of expertise as a company, right? Like the version control you use doesn't matter with Coder. There's a tighter integration with 3 major git cloud platforms, but you could use SVN or Perforce too.

Open to thoughts on this, and I like the idea of a doc, but I'm curious how much we should maintain versus acknowledging that `git` knowledge is outside the scope of our product in a general sense.

guides/gitconfig.md
IAmHughes: I wonder if we can put somewhere in this doc the fact that Coder, by default, will load a `.gitconfig` using your name and email associated with your Coder login. That was a point for me because I was forced to use the dotfiles setup to ensure my `gitconfig` had the proper email address to keep my commits anonymous on GitHub. GitHub allows for private emails, and therefore automatically adding one's email to a gitconfig where they can push to public repositories could expose their email.

We call this out below, but I feel like there's room for highlighting this.

guides/gitconfig.md
ericpaulsen: Agree as well - will change to something like:

> Below is an example of `.gitconfig` contents that could be applied to `/etc/gitconfig`. Here we are making `ssh` the System-level default when engaging with Git providers.

```gitconfig
[url "git@github.com:"]
    insteadOf = https://github.com/
```

Let me know if this works or if I'm off-base here.

IAmHughes (dismissed):

guides/gitconfig.md
ericpaulsen: > Do we want to provide examples? There are a lot of guides out there on `gitconfig` and while we can provide one too, it's not necessarily our area of expertise as a company, right? Like the version control you use doesn't matter with Coder. There's a tighter integration with 3 major git cloud platforms, but you could use SVN or Perforce too.
> 
> Open to thoughts on this, and I like the idea of a doc, but I'm curious how much we should maintain versus acknowledging that `git` knowledge is outside the scope of our product in a general sense.

Understand your point here, but the purpose is to provide context on `.gitconfig` as it relates to Coder & how users can manage their git configs thought the various levels (system, global, etc.)

guides/gitconfig.md
ericpaulsen: Agree - I'm happy to highlight this further. Any thoughts on how we should go about that?

guides/gitconfig.md
IAmHughes: Right, understandable. I don't think it's a bad thing to provide info on, but `--system`, `--local`, `--global` aren't unique to Coder was my point.

vapurrmaid: Given that we inject a default `.gitconfig` and that we adjusted build order based off of end-user feedback, I think an example could be useful but I'm also not against scrapping the example.

I'll leave that decision to y'all.

My only point is that if we keep the example in, we change the code fences, mention it's example contents of a `.gitconfig` for using `ssh` as default within an example Coder environment. We could even say we ourselves use this setting if we feel it's beneficial to mention so.

OTherwise, ya, I'll defer to @IAmHughes @khorne3  and @ericpaulsen

IAmHughes: > My only point is that if we keep the example in, we change the code fences, mention it's example contents of a .gitconfig for using ssh as default within an example Coder environment. We could even say we ourselves use this setting if we feel it's beneficial to mention so.

Totally on board with that approach!

guides/gitconfig.md
vapurrmaid: I was wrong about this comment - the recommended way to add a system gitconfig would be directly in the Dockerfile:

```dockerfile
# Add system level gitconfig
COPY [".gitconfig", "/etc/gitconfig"]
```

In fact, we do this ourselves.

I'm sorry for steering in the wrong direction with `coder/configure`, which is for more complicated needs.

For more information on `coder/configure`, we can easily refer to <https://coder.com/docs/images/configure>

ericpaulsen: No worries - appreciate the clarification. I'll push another commit to clarify this and add the manifest file.

guides/gitconfig.md
```diff
@@ -0,0 +1,32 @@
...
+global, and repo levels.
+
+## How to use the System and Global configurations
+
+For scenarios whereby a set of git configurations should apply across an organization,
+we recommend using a System-level git configuration. System-level git configurations
+are located at `/etc/gitconfig`. These settings are applied to each git repository,
+but may be overriden using global or worktree git configurations.
+
+We suggest adding the System-level `.gitconfig` directly in the image Dockerfile.
```
ericpaulsen, line 16: let me know if "image Dockerfile" is correct. it sounds a bit off but I think it's the right way to say it.

guides/gitconfig.md
vapurrmaid: Trivial suggestion - maybe the first section of the article can be about our `gitconfig` injection?

ericpaulsen: Good idea - we lead with the default functionality, and then guide the customer through overriding such functionality.

guides/gitconfig.md
khorne3: I searched to see how Docker words it and they use "image's Dockerfile"

khorne3 (approved):

guides/gitconfig.md
```diff
@@ -0,0 +1,36 @@
+---
+title: Managing Git Configuration
+description: Learn how to configure Git in Coder.
+---
+
+This guide will show you how to manage your Git configuration in Coder.
+
+## Personal git Configurations
+
+By default, Coder injects a .gitconfig file into each environment if there isn't
```
vapurrmaid, line 10:
```suggestion
By default, Coder injects a `.gitconfig` file into each environment if there isn't
```

guides/gitconfig.md
```diff
@@ -0,0 +1,36 @@
...
+a system-level or a user-defined .gitconfig file; the file sets the user name
```
vapurrmaid, line 11: We need to adjust this - it's not quite accurate. We look for a global `.gitconfig` located at `~/.gitconfig`. Whether or not there's a system `/etc/gitconfig` is irrelevant.

Bare with me for a second here:

1. The build order is as follows: `git configuration`,  `coder/configure`, `personalize`. You can see this from a build log order:

Screenshot from 2021-03-04 13-47-21

2. The `git configuration` steps looks for `~/.gitconfig`
3. If no `~/.gitconfig` is found, it creates a default one using your email address and name from Coder

The line of code is literally:

```bash
if [[ -f $GLOBAL_GIT_CONFIG_PATH ]]; then
    return 0
fi
```

What this implies is we will always create one for you when you create a new environment, but it may be replaced in `coder/configure` or `personalize`, and it won't be re-injected on rebuilds if a file is already located at `~/.gitconfig`.

guides/gitconfig.md
```diff
@@ -0,0 +1,36 @@
...
+configuration. We suggest adding the System-level `.gitconfig` directly to the
+image's Dockerfile:
+
+```Dockerfile
+# Add system-level gitconfig
+COPY ["gitconfig", "/etc/gitconfig"]
+```
+
+As you can see from the example, Coder stores system-level git configurations
+under **/etc/gitconfig**. If present, Coder applies the settings to each git
```
vapurrmaid, line 34: This is not `Coder` specific. I wouldn't mention `Coder stores` at all here. This is `git` concept and can be referenced from:

https://git-scm.com/docs/git-config#Documentation/git-config.txt---system

and is generally true when git is installed in a unix-like environment

guides/gitconfig.md
ericpaulsen: I see - so rather than saying: 

>if there isn't a system-level or user-defined `.gitconfig`

it should really say:

> Coder will always inject a `.gitconfig` file in a newly created environment, but it may be overwritten in `coder/configure` or `~/personalize`, and won't be re-injected on rebuilds if a file is already located at `~/.gitconfig`.

Let me know what you think.

vapurrmaid: Yea, something like that. How's?

> Coder will create a global git configuration file located at `~/.gitconfig` in all newly created environments.
> This step occurs before `coder/configre` (add a link) and `personalization` (add a link), which may be utilized
> to replace the default `~/.gitconfig` created by Coder. Whenever an environment is rebuilt, we'll skip creating
> a default `~/.gitconfig` if the file already exists.

What's missing is that it contains Coder email address and name - those need to be added.

CC @khorne3 for improved product copy

guides/gitconfig.md
ericpaulsen: So to clarify:

> As you can see from the example, system-level git configurations live under `/etc/gitconfig`.

PR #141, opened 2021-03-04
ericpaulsen/gitconfig

ericpaulsen:
new PR to reflect the most recent commit based on previous feedback on closed [PR 138](https://coder.com/docs/images/configure)


khorne3 (approved):

guides/gitconfig.md
```diff
@@ -0,0 +1,41 @@
...
+
+```Dockerfile
+# Add system-level gitconfig
+COPY ["gitconfig", "/etc/gitconfig"]
+```
+
+As you can see from the example, system-level git configurations live
+under `/etc/gitconfig`. If present, Coder applies the settings to each git
+repository present. However, any Coder user can override system-level settings
+using global or worktree git configurations.
```
vapurrmaid, line 41: **suggestion** Something alone the lines of:

```
If present, they're applied to each git repository. However, any Coder user can override
system-level settings using global or worktree git configurations
```

As it's not `Coder` applying any of these settings. It's how `git` works on a unix-like system.

vapurrmaid (changes requested):
Other than the one requested change, I think it would be worthwhile adding at the bottom of the document something along the lines of:


For more information on git configuration, refer to the official documentation: <https://git-scm.com/docs/git-config>

guides/gitconfig.md
khorne3: Small change: 

> If present, the git applies the settings to each repository. However, any Coder user can override
system-level settings using global or worktree git configurations.

Otherwise we're good to go

vapurrmaid: I think it should be `git` instead of `the git` :P

IAmHughes (approved):
Approving, pending the changes from @vapurrmaid and @khorne3 's [discussion](https://github.com/cdr/docs/pull/141/files#r587867929)

guides/gitconfig.md
ericpaulsen: @vapurrmaid - I pushed a new commit with the changes discussed above, so we should be good to merge. but before I do, I'd appreciate if you could take a look, just to avoid another revert 😄

IAmHughes (approved):
LGTM, I think this covers what @vapurrmaid asked for as well.

PR #143, opened 2021-03-05
ericpaulsen/fixgkecommand

ericpaulsen:
GKE create cluster command wasn't working properly, due to env variable syntax being incorrect. PR includes fix & adding of `$ZONE` variable, in the place of the previous `us-central1-a` zone (which won't necessarily apply to each customer).


ericpaulsen:
accidentally included all commits from gitconfig branch into this one. closing this one out and submitting a new one with the correct commits.

PR #146, opened 2021-03-05
fix: home directory

ericpaulsen:
updating from `home/<username>` to `home/coder`

### DIFF
--- a/environments/lifecycle.md
+++ b/environments/lifecycle.md
@@ -12,8 +12,8 @@ on cloud compute and justify more powerful dev machines :).
 Rebuilding an [Environment](index.md) allows you to update to the latest image,
 edit resource requests, or restart your environment after a shutdown.
 
-Only the `/home/<username>` directory persists between rebuilds. Rebuilds
-do not affect configurations and source code within the `/home/<username>`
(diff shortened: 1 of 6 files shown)


cmoog:
Isn't the existing language is correct? Different images have different users.

ericpaulsen:
> Isn't the existing language is correct? Different images have different users.

Ah you're right - I thought we used the user's Coder username as part of the home directory in previous versions, so I thought it was outdated. Makes sense that it is set on the image. Should we clarify that `<username` is set in the image?

vapurrmaid:
> > Isn't the existing language is correct? Different images have different users.
> 
> Ah you're right - I thought we used the user's Coder username as part of the home directory in previous versions, so I thought it was outdated. Makes sense that it is set on the image. Should we clarify that `<username>` is set in the image?

I think we can mention it. In our development image, the last line is:

```dockerfile
USER coder
```

For example in the below section: https://coder.com/docs/environments/lifecycle#rebuilds

We could have a 

> **Note**: `username` is defined in an image. See <https://docs.docker.com/engine/reference/builder/#user>

cmoog:
Oh I see where confusion is coming from. Yeah that all sounds good to me.

vapurrmaid (approved):

PR #147, opened 2021-03-08
Chore: Clarify updating steps to include values override

IAmHughes:
This PR clarifies the `updating` steps by being explicit in ensuring you retain your current values in your helm installation.

Change visible in preview deployment: 

image

### DIFF
--- a/setup/updating.md
+++ b/setup/updating.md
@@ -37,12 +37,20 @@ Updating Coder is a two-step process:
     ```bash
     helm repo update
     ```
+    
+2. Export the current helm chart values into a file:
 
-2. Upgrade to the desired version (i.e., `1.16.0`):
+    ```bash
+    helm get values --namespace coder coder > current-values.yml
+    ```
+
+3. Upgrade to the desired version with your helm chart values file (i.e., `1.16.0`):
+
+    _Note: Omitting `--version` will default to the latest version_
 
     ```bash
     helm upgrade --namespace coder --force --install --atomic --wait \
-      --version 1.16.0 coder coder/coder
+      --version 1.16.0 coder coder/coder -f current-values.yml
     ```
 
 ## Fixing a Failed Upgrade
--- a/setup/updating.md
+++ b/setup/updating.md
@@ -47,7 +47,7 @@ To update Coder, follow these steps:
 1. Provide your helm chart values file and upgrade to the desired version (e.g.,
 1.16.1):
 
-    Note: If you omit --version, you'll upgrade to the latest version.
+    *Note: If you omit --version, you'll upgrade to the latest version.*
 
     ```bash
     helm upgrade --namespace coder --force --install --atomic --wait \
--- a/setup/updating.md
+++ b/setup/updating.md
@@ -30,7 +30,7 @@ which you've deployed Coder.
 
 ## Update Coder
 
-Updating Coder is a two-step process:
+To update Coder, follow these steps:
 
 1. Retrieve the latest repository information:
 
@@ -38,19 +38,19 @@ Updating Coder is a two-step process:
     helm repo update
     ```
 
-2. Export the current helm chart values into a file:
+1. (Optional) Export the current helm chart values into a file:
 
     ```bash
     helm get values --namespace coder coder > current-values.yml
     ```
 
-3. Upgrade to the desired version with your helm chart values file (i.e., `1.16.0`):
+1. Upgrade to the desired version with your helm chart values file (i.e., `1.16.1`):
 
     _Note: Omitting `--version` will default to the latest version_
 
     ```bash
     helm upgrade --namespace coder --force --install --atomic --wait \
-      --version 1.16.0 coder coder/coder -f current-values.yml
+      --version 1.16.1 coder coder/coder --values current-values.yml
     ```
 
 ## Fixing a Failed Upgrade
@@ -84,7 +84,7 @@ If this happens, we recommend uninstalling and reinstalling:
     helm get values --namespace coder coder > current-values.yml
     ```
 
-2. Run `helm uninstall`:
+1. Run `helm uninstall`:
 
     ```bash
     helm uninstall --namespace coder coder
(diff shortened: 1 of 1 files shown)


setup/updating.md
```diff
@@ -37,12 +37,20 @@ Updating Coder is a two-step process:
     ```bash
     helm repo update
     ```
+    
+2. Export the current helm chart values into a file:
```
jawnsy, line 41: I suggest numbering these with 1. 1. 1. so that we can get autonumbering. Also the above text ("Updating Coder is a two-step process") is now inaccurate :wink:

Perhaps we can rephrase to avoid mentioning a number of steps, for future updates, maybe: "To update Coder, follow these steps:" (I am sure @khorne3 can phrase it better though)

setup/updating.md
```diff
@@ -37,12 +37,20 @@ Updating Coder is a two-step process:
...
 
-2. Upgrade to the desired version (i.e., `1.16.0`):
+    ```bash
```
jawnsy, line 43: I think we should use `console` instead of `bash` as these are commands that someone is typing in manually

https://github.com/github/linguist/blob/8624ea022d3c78a5e8687efede4ea64a8b5cae1e/lib/linguist/languages.yml#L5314-L5325

Here's what this looks like with bash:

```bash
helm get values --namespace coder coder > current-values.yml
```

And the same with console:

```console
$ helm get values --namespace coder coder > current-values.yml
```

(I like to use `$` to show that this can be run as a non-root user)

All that said, I understand we might not want to change this now, because we would want things to be consistent across the docs... so leaving as `bash` is fine for now

setup/updating.md
```diff
@@ -37,12 +37,20 @@ Updating Coder is a two-step process:
...
+    helm get values --namespace coder coder > current-values.yml
```
jawnsy, line 44: suggest making this an optional step, the preferred flow (IMO) should be that an admin already has their values file (because they version controlled it), so this step is only needed if they _gasp_ forgot to do that when they installed/etc

setup/updating.md
```diff
@@ -37,12 +37,20 @@ Updating Coder is a two-step process:
...
+    ```
+
+3. Upgrade to the desired version with your helm chart values file (i.e., `1.16.0`):
+
+    _Note: Omitting `--version` will default to the latest version_
 
     ```bash
     helm upgrade --namespace coder --force --install --atomic --wait \
-      --version 1.16.0 coder coder/coder
+      --version 1.16.0 coder coder/coder -f current-values.yml
```
jawnsy, line 53: version should be 1.16.1 now

Also my preference is to use `--values` instead of `-f` (long commands are always preferable to make what the commands are doing more obvious)

jawnsy (approved):
LGTM, thanks for updating it! :+1:

setup/updating.md
```diff
@@ -38,11 +38,19 @@ Updating Coder is a two-step process:
...
-2. Upgrade to the desired version (i.e., `1.16.0`):
+2. Export the current helm chart values into a file:
+
+    ```bash
+    helm get values --namespace coder coder > current-values.yml
+    ```
+
+3. Upgrade to the desired version with your helm chart values file (i.e., `1.16.0`):
+
+    _Note: Omitting `--version` will default to the latest version_
```
jawnsy, line 49: should clarify whether it's the latest release version or latest release candidate, I think the former but it seems worth double checking

setup/updating.md
IAmHughes: Agreed with the wording. I almost did the `1.` thing instead actually out of habit but saw the list was actually numbered so I figured I'd keep it, haha. Updating it though because I prefer it too 😋

setup/updating.md
IAmHughes: good point!

setup/updating.md
IAmHughes: Yeah, we need to consider how we want to handle versions probably @khorne3. Updating the docs for every single patch release is going to be a pain in terms of technical debt.

Agreed on `--values` which means I'll update the doc section below this too

setup/updating.md
IAmHughes: @khorne3 this is a good note for the future as well. We could go through and change all of them if we wanted, but that's a bit outside the scope of this PR. Perhaps moving forward we can update them all? Let me know your thoughts.

setup/updating.md
IAmHughes: @jawnsy who manages our helm repo? @deansheather ? Not sure how to know if this defaults to the latest release candidate or not, but I assume so?

setup/updating.md
khorne3: @khorne3 I like this idea. I created a [CH card](https://app.clubhouse.io/coder/story/8969/change-code-fencing-from-bash-to-console) so this can be tracked (definitely think it should all be done at once)

setup/updating.md
```diff
@@ -30,19 +30,27 @@ which you've deployed Coder.
...
-2. Upgrade to the desired version (i.e., `1.16.0`):
+1. (Optional) Export the current helm chart values into a file:
+
+    ```bash
+    helm get values --namespace coder coder > current-values.yml
+    ```
+
+1. Upgrade to the desired version with your helm chart values file (i.e., `1.16.1`):
+
+    _Note: Omitting `--version` will default to the latest version_
```
khorne3, line 49: I thought the first line was a bit confusing, so maybe

> Provide your helm chart values file and upgrade to the desired version (e.g.,
`1.16.1`):
> 
> *Note: If you omit --version, you'll upgrade to the latest version.*

IAmHughes: I like it, updating!

setup/updating.md
```diff
@@ -30,19 +30,28 @@ which you've deployed Coder.
...
+1. (Optional) Export the current helm chart values into a file:
+
+    ```bash
+    helm get values --namespace coder coder > current-values.yml
+    ```
+
+Provide your helm chart values file and upgrade to the desired version (e.g.,
+1.16.1):
+
+Note: If you omit --version, you'll upgrade to the latest version.
```
IAmHughes, line 50: @khorne3 how's that look?

khorne3 (approved):

setup/updating.md
IAmHughes: Merging as is now with approvals, but would love to hear some input on this for a future change to the doc.

PR #162, opened 2021-03-15
chore: add notes re: update process

khorne3:
Getting the conversation started re: backups, number of versions to upgrade, etc.

### DIFF
--- a/setup/updating.md
+++ b/setup/updating.md
@@ -30,15 +30,11 @@ which you've deployed Coder.
 
 ## Recommendations
 
-- As with any significant maintenance operation, we recommend taking a snapshot of
-the database before proceeding with the upgrade.
-
-    Although Coder tests in-place upgrades and rollbacks involving database
-    state, it is good practice to have a recent backup created during a
-    maintenance window immediately prior to the upgrade. It is simpler and safer
-    to roll back directly at the database level, and Coder recommends doing so
-    in the event of an upgrade failure, since it guarantees restoration of the
-    system to a known-working condition.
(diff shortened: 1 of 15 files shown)


f0ssel (approved):
LGTM

setup/updating.md
```diff
@@ -28,6 +28,13 @@ helm repo add coder https://helm.coder.com
 Please also ensure that your Kubernetes config is pointing to the cluster on
 which you've deployed Coder.
 
+## Recommendations
+
+- Before beginning, we **strongly recommend** taking snapshots
```
jawnsy, line 33: I would suggest a phrasing like:

> As with any significant maintenance operation, we recommend taking a snapshot of the database before proceeding with the upgrade. Although Coder tests in-place upgrades and rollbacks involving database state, it is always a good practice to have a recent backup created during a maintenance window immediately prior to the upgrade. It is simpler and safer to roll back directly at the database level, and Coder recommends doing so in the event of an upgrade failure, since it guarantees restoration of the system to a known-working condition.

@coadler is working on our rollback/reversion strategy so may have some more to add in that regard

setup/updating.md
```diff
@@ -28,6 +28,13 @@ helm repo add coder https://helm.coder.com
...
+of your environment in case there are issues with the update process.
+- We recommend updating no more than version at a time (e.g., we recommend
```
jawnsy, line 35: I'm curious how much of a jump we want to support, whether we're talking about 1.15 -> 1.16 or also including minor revisions?

deansheather (approved):

PR #168, opened 2021-03-16
chore: Add local lint scripts

vapurrmaid:
Resolves: #112 

## Context

We have linting enabled (via markdownlint) in CI, but no local means of
running lint checks.

## Summary

Install markdownlint-cli locally and add package.json scripts for
running lint commands


package.json
```diff
@@ -1,5 +1,10 @@
 {
+  "scripts": {
+    "lint": "markdownlint '**/*.md'",
```
jawnsy, line 3: I'm not sure if this is too hard to do, but it would be nice to change CI to use these scripts, so that we get testing for these lines?

jawnsy (approved):

package.json
vapurrmaid: Agreed!

vapurrmaid: Also the fact that the lint tool is installed on each run (minus any caching) means we could potentially run into a version mismatch and therefore a mismatch in output

vapurrmaid: Gotta run to a meeting...started what I think is the right steps here: https://github.com/cdr/docs/pull/168/commits/3adf46fb1eca2e7f7acbc25b632f189eb5aaae1d

.github/workflows/ci.yaml
```diff
@@ -10,5 +10,7 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: 12.x
-      - run: npm install -g markdownlint-cli
-      - run: markdownlint ./
\ No newline at end of file
+      - name: Install node_modules
+        run: yarn
```
jawnsy, line 14: Nice, does this mean that the version of markdownlint is now versioned according to devDependencies? (I'm not too familiar with how node/npm/yarn handle this)

If so, sweet!

jawnsy (approved):
The new steps are running in CI, seems good to me!

.github/workflows/ci.yaml
vapurrmaid: Yep!

PR #179, opened 2021-03-19
chore: add doc on increasing inotify-watch-limits

khorne3:
For [CH#6016](https://app.clubhouse.io/coder/story/4988/document-increasing-inotify-watcher-limit-at-the-node-level)


guides/inotify-watch-limits.md
```diff
@@ -0,0 +1,144 @@
...
+              cpu: 0.01
+          image: alpine:3.6
+          name: sleepforever
+          command: ["tail"]
+          args: ["-f", "/dev/null"]
+      volumes:
+        - name: sys
+          hostPath:
+            path: /sys
+```
```
scsmithr, line 71: I think we're safe with deleting this and just showing the one below. I don't know why one would need to set `apiVersion: extensions/v1beta1` over `apiVersion: apps/v1`

guides/inotify-watch-limits.md
```diff
@@ -0,0 +1,96 @@
...
+    ```console
+    kubectl delete more-fs-watchers
+    ```
+
+    This command removes all related daemonset pods, and no further pods will be
+    spun up.
+
+## Helpful Resources
+
+- [Setting Sysctls for a
```
jawnsy, line 93: I don't think this actually helps in this case, because these sysctls are not namespaced and therefore not on the allow list

guides/inotify-watch-limits.md
```diff
@@ -0,0 +1,96 @@
+---
+title: Inotify Watch Limit Increase
```
jawnsy, line 2: Suggest titling this "Increasing the inotify Watch Limit" or "Increasing the File Watcher Limit" (I think [inotify_add_watch](https://linux.die.net/man/2/inotify_add_watch) is typically styled in lowercase as it's the name of a system call)

```suggestion
title: Increasing the inotify Watch Limit
```

More generally, though, it seems we may want to have this be part of a larger page about tuning options (or more specifically, kernel tuning options) for Coder?

guides/inotify-watch-limits.md
```diff
@@ -0,0 +1,96 @@
...
+description: Learn how to increase the inotify watcher limit.
+---
+
+When running a webpack project, you may encounter an error similar to the
```
jawnsy, line 6: This is not Webpack-specific, so I would suggest making this more generic.

```suggestion
With some applications and tools, including Webpack or code-server, you may encounter an error similar to the
```

bpmct:
As we get the ability to group guides together, could we put this into something like `troubleshooting`?

khorne3:
> As we get the ability to group guides together, could we put this into something like `troubleshooting`?

@bpmct Yep! I spun up a [working preview](https://github.com/cdr/docs/pull/180) so I'll add a troubleshooting section too

guides/inotify-watch-limits.md
```diff
@@ -0,0 +1,96 @@
...
+following:
+
+```text
+Watchpack Error (watcher): Error: ENOSPC: System limit for number of file
+watchers reached, watch '/some/path'
+```
+
+This results from a low number of [inotify
```
jawnsy, line 14: Suggest adding more info about what the inotify mechanism is for, what it does, and the impact of changing the watch limit.

```suggestion
The [inotify facility](https://en.wikipedia.org/wiki/Inotify) allows programs to monitor files and directories for changes, so that they will receive an event immediately whenever a user or program modifies the file or directory. Because the inotify mechanism requires some kernel resources (memory and processor) per watched file, there is a limit on the number of registered file watchers. For large projects with many files and folders, programs may hit this limit, resulting in the above error message.
```

guides/inotify-watch-limits.md
```diff
@@ -0,0 +1,96 @@
...
+watches](https://confluence.jetbrains.com/display/IDEADEV/Inotify+Watches+Limit)
+combined with high node usage, causing the log stream to fail.
+
+## Resolution
+
+Increase the number of file watchers. Because the setting quantifying the number
```
jawnsy, line 20:
```suggestion
If you encounter the file watcher limit, there are two potential resolutions: either reduce the number of file watcher registrations, or increase the maximum file watcher limit. We recommend attempting to reduce the file watcher registrations first, because increasing the number of file watches may result in high processor utilization.

For code-server, you may exclude some files, such as those belonging to third-party project dependencies, by modifying the `files.watcherExclude` setting. Please see the user guides for the software registering the watches to learn more.
```

guides/inotify-watch-limits.md
```diff
@@ -0,0 +1,96 @@
...
+of file watchers isn't namespaced, you'll need to raise the maximum number at
+the node level.
+
+One way to do this is to use a daemonset in the cluster with a privileged
```
jawnsy, line 24:
```suggestion
One way to do this is to use a DaemonSet in the cluster with a privileged
```

DaemonSet is a proper noun so should be capitalized, and we should stylize it as [indicated in the official Kubernetes documentation](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/)

guides/inotify-watch-limits.md
```diff
@@ -0,0 +1,96 @@
...
+      labels:
+        name: more-fs-watchers
+        k8s-app: more-fs-watchers
+    spec:
+      hostNetwork: true
+      hostPID: true
+      hostIPC: true
+      initContainers:
+        - command:
+            - sh
```
jawnsy, line 51: @scsmithr How come this uses `sh -c` instead of directly running the `sysctl` command?

guides/inotify-watch-limits.md
```diff
@@ -0,0 +1,96 @@
...
+          name: sysctl
+          resources: {}
+          securityContext:
+            privileged: true
+          volumeMounts:
+            - name: sys
+              mountPath: /sys
+      containers:
+        - resources:
+            requests:
```
jawnsy, line 65: @scsmithr we should set requests/limits too. The pod should be set with a read-only rootfs for security. And I suggest we use the sleep command instead of tail -f?

I am actually not sure if we need privileges other than the ability to write to sysfs? We should check this...

guides/inotify-watch-limits.md
```diff
@@ -0,0 +1,96 @@
...
+- You can **delete** this by running:
+
+    ```console
+    kubectl delete more-fs-watchers
+    ```
+
+    This command removes all related daemonset pods, and no further pods will be
+    spun up.
+
+## Helpful Resources
```
jawnsy, line 91:
```suggestion
## See Also
```

We can't guarantee these will be helpful :wink:

guides/inotify-watch-limits.md
```diff
@@ -0,0 +1,96 @@
...
+  selector:
+    matchLabels:
+      k8s-app: more-fs-watchers
+  template:
+    metadata:
+      labels:
+        name: more-fs-watchers
+        k8s-app: more-fs-watchers
+    spec:
+      hostNetwork: true
```
jawnsy, line 46: @scsmithr Do these really need to share host namespaces? I don't believe so, since sysfs is being mounted read-write. We shouldn't recommend things that don't meet Kubernetes' security recommendations in our docs

Especially since this is privileged and running as a DaemonSet, we should try to restrict things as much as possible: https://kubernetes.io/docs/concepts/security/pod-security-standards/

guides/inotify-watch-limits.md
jawnsy: I tested this and it works - there's only so much we can do since we need a privileged container to access /proc/sys, but I at least tried to cut down on the host namespaces we were sharing into the DaemonSet pods:

```yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: more-fs-watchers
  namespace: kube-system
  labels:
    app: more-fs-watchers
    k8s-app: more-fs-watchers
spec:
  selector:
    matchLabels:
      k8s-app: more-fs-watchers
  template:
    metadata:
      labels:
        name: more-fs-watchers
        k8s-app: more-fs-watchers
    spec:
      initContainers:
        - name: sysctl
          image: alpine:3
          command:
            - sysctl
            - -w
            - fs.inotify.max_user_watches=16384
          resources:
            requests:
              cpu: 10m
              memory: 1Mi
            limits:
              cpu: 100m
              memory: 5Mi
          securityContext:
            runAsUser: 0
            privileged: true
            readOnlyRootFilesystem: true
      containers:
        - name: pause
          image: gcr.io/google_containers/pause-amd64:3.1
          command:
            - /pause
          resources:
            requests:
              cpu: 10m
              memory: 1Mi
            limits:
              cpu: 100m
              memory: 5Mi
          securityContext:
            allowPrivilegeEscalation: false
            runAsUser: 1000
            runAsNonRoot: true
            readOnlyRootFilesystem: true
```

jawnsy:
I've opened #183 and will work on this there

PR #185, opened 2021-03-22
chore: add superuser requirement

khorne3:
For [CH#2927](https://app.clubhouse.io/coder/story/2927/document-database-superuser-requirement)


khorne3:
@cmoog Is this only relevant here, or is it necessary for initial setup too?

cmoog:
It applies in both cases. Whenever the product deploys we run database migrations, which in both cases may need to modify the database schema (which is what requires those perms).

cmoog (approved):

PR #199, opened 2021-03-23
fix(docs): Consolidate Dev URL(s) naming

vapurrmaid:
We refer to Developer URLs as two words, both capitalized: `Dev URL(s)`. I found various cases where the two words were combined into one, which is inconsistent with most other cases and the product UI.

### DIFF
--- a/admin/workspace-providers/deployment.md
+++ b/admin/workspace-providers/deployment.md
@@ -116,7 +116,7 @@ when communicating with the Coder deployment.
    Optionally, you can provide additional helm configuration values by providing
    a `values.yaml` file and adding the argument `-f my-values.yaml` to the
    generated command. Helm values control attributes of the workspace provider,
-   including DevURLs, Kubernetes storage classes, SSH, and more. See the
+   including Dev URLs, Kubernetes storage classes, SSH, and more. See the
    [Workspace Provider Helm Chart
    Values]("https://github.com/cdr/enterprise-helm/blob/workspace-providers-envproxy-only/README.md")
    for more details.
--- a/changelog/1.13.2.md
+++ b/changelog/1.13.2.md
@@ -9,7 +9,7 @@ There are no breaking changes in 1.13.2.
 
 ### Features ✨
 
-- infra: Adds support for `https` servers behind DevURLs
+- infra: Adds support for `https` servers behind Dev URLs
   - **Notice:** Environments will need to be rebuilt before using this feature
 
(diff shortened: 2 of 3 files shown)


admin/workspace-providers/deployment.md
```diff
@@ -116,7 +116,7 @@ when communicating with the Coder deployment.
    Optionally, you can provide additional helm configuration values by providing
    a `values.yaml` file and adding the argument `-f my-values.yaml` to the
    generated command. Helm values control attributes of the workspace provider,
-   including DevURLs, Kubernetes storage classes, SSH, and more. See the
+   including Dev URLs, Kubernetes storage classes, SSH, and more. See the
```
jawnsy, line 119: Ah, we need to apply this change to the rest of our product too, then... :grimacing: 

I think [Storage Classes should be capitalized](https://kubernetes.io/docs/concepts/storage/storage-classes/) because it's a proper noun. Kubernetes seems inconsistent with how they use spacing (e.g. [DaemonSet](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/) and [ReplicaSet](https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/) are one word in the docs, and it matches the Kubernetes object `kind`, but Storage Classes is two words and does not match the Kubernetes object `kind`)

What do you think about cross-linking between the relevant topics when possible? It helps someone reading the doc to know where to learn more about some concept, in case it's their first time encountering it (depending on the ordering that they're reading the docs)

jawnsy (approved):

admin/workspace-providers/deployment.md
vapurrmaid: > Ah, we need to apply this change to the rest of our product too, then...

See https://github.com/cdr/m/commit/c154e7f6bb1fc7575d9f30c42f63244d49cfa1ce

cross-linking is a +1 IMO, and there are tools to check if the links start to 404. We could do this in a cron or some other CI workflow as appropriate: https://github.com/tcort/markdown-link-check

jawnsy: Already ahead of you grey! https://app.clubhouse.io/coder/story/9656/check-for-broken-links-as-part-of-ci-build

vapurrmaid: Sounds good @jawnsy . I'm going to go ahead and merge this. Feel free to open another PR against the consistencies for Kubernetes nouns!

Issue #63, opened 2021-01-26
Helper Scripts for Changelogs


vapurrmaid:
Generate `x.y.z.md` from a template

PR #200, opened 2021-03-23
chore: Generate clog from template

vapurrmaid:
Resolves: #63

## Summary

A simplistic shell script for creating new changelogs from a template

## Details

Adds `scripts/create-clog.sh <version> <release-date>` utility.

## Note

I did not try to use `jq` or some other means to add the changelog to `manifest.json`. I started down that path, but there were some quirks (example sorting) and writing back to the file properly, that I feel it wasn't worth the effort for the time being.


scripts/template_changelog.md
```diff
@@ -0,0 +1,20 @@
+---
+title: "<% X.Y.Z %>"
+description: "Released on <% MM/DD/YYYY %>"
+---
+
+### Breaking Changes ❗
```
jawnsy, line 6: I love the cutesy emojis btw

@khorne3 mentioned we should use sentence case for lower-level headings, so I think this would be "Breaking changes" ?

scripts/create-clog.sh
```diff
@@ -0,0 +1,72 @@
...
+VERSION_DELIM="<% X.Y.Z %>"
+RELEASE_DATE_DELIM="<% MM/DD/YYYY %>"
+VERSION=""
+RELEASE_DATE=""
+
+# usage prints information for how to use this program
+usage () {
+  echo "Usage: create-clog <version> <release-date>"
+  echo "Create a changelog from a template"
+  echo "Arugments:"
```
jawnsy, line 24:
```suggestion
  echo "Arguments:"
```

Not sure if you meant Arugula :wink:

scripts/create-clog.sh
```diff
@@ -0,0 +1,72 @@
...
+
+set -eou pipefail
+
+VERSION_DELIM="<% X.Y.Z %>"
+RELEASE_DATE_DELIM="<% MM/DD/YYYY %>"
+VERSION=""
+RELEASE_DATE=""
+
+# usage prints information for how to use this program
+usage () {
```
jawnsy, line 21:
```suggestion
function usage () {
```

just a personal preference but for bash-only scripts, the function keyword makes it clearer that this is a function: https://stackoverflow.com/questions/7917018/what-is-the-function-keyword-used-in-some-bash-scripts

we try to use that consistently in other parts of our bash scripts

scripts/create-clog.sh
```diff
@@ -0,0 +1,72 @@
...
+  echo "Usage: create-clog <version> <release-date>"
+  echo "Create a changelog from a template"
+  echo "Arugments:"
+  echo "    <version>: A version string 'x.y.z'"
+  echo "    <release-date>: A date string 'mm/dd/yyyy'"
+}
+
+# init parses arguments and initializes all variables.
+init () {
+  lenArgs=2
```
jawnsy, line 31: consider getopt for this! it's actually not that bad to use, we use it in many of our scripts now...

https://dustymabe.com/2013/05/17/easy-getopt-for-a-bash-script/

scripts/create-clog.sh
```diff
@@ -0,0 +1,72 @@
...
+  fi
+  
+  if [ -z "${2-}" ]; then
+    echo "release-date argument is empty."
+    echo "Expected a date string like 'mm/dd/yyyy'"
+    usage
+    exit 1
+  fi
+  
+  VERSION="${1}"
```
jawnsy, line 53: @deansheather pointed out that `$2` is easier to read, so we should only use `{}` when necessary, this applies to `$VERSION` and `$RELEASE_DATE` too

scripts/create-clog.sh
```diff
@@ -0,0 +1,72 @@
...
+  RELEASE_DATE="${2}"
+}
+
+# createFromTemplate creates a new changelog file from template_changelog.md
+# using VERSION and RELEASE_DATE.
+createFromTemplate () {
```
jawnsy, line 59: I think snake_case is more common with bash scripts

jawnsy (approved):

scripts/template_changelog.md
vapurrmaid: If we do that, we should update all clogs. Happy for us to do that, though as a different changeset. I'm super glad we're increasing our consistency and standardizing these decisions!!!!! :heart:

scripts/create-clog.sh
vapurrmaid: I just ate some :green_salad: (albeit in smoothie form)

scripts/create-clog.sh
vapurrmaid: Ah, yes - great call

scripts/create-clog.sh
jawnsy: also totally unrelated but I started writing unit tests for our bash scripts using BATS in the monorepo, lol

scripts/create-clog.sh
vapurrmaid: Nice, TIL

vapurrmaid: Do you have a useful summary or reference of when we must use `{}`. We used to always write our scripts that way, so I sort of reach for it reflexively. Again, I'm super happy with the way we're consolidating and standardizing our usage. Thank you!

scripts/template_changelog.md
jawnsy: btw, and it pains me a bit to say this, because I love the emojis, but... the emojis result in slightly worse fragments: `https://coder.com/docs/changelog/1.17.0#breaking-changes-` (note the trailing hyphen)

I dunno if there's some way to customize this in the JavaScript we're using to transform the Markdown, though

scripts/create-clog.sh
jawnsy: I think `{}` is needed if you're using the default-value syntax `${VERSION:-abc}` to use `$VERSION` if specified, or fall back to `abc` as a default value, or if the text adjacent to your variable contains an underscore or is just letters (e.g. `echo "This is version ${VERSION}_beta"` or `echo "This is version ${VERSION}rc1"`).

In those cases if you're using `set -euo pipefail` as we should be, then you'll just get an error about an undefined variable, so then you know to add the `{}`.

jawnsy:
Webpack seems displeased with your addition! From the Vercel logs (just a warning though):

```
12:21:48.391 | ./docs/scripts/create-clog.sh 1:0
12:21:48.391 | Module parse failed: Unexpected character '#' (1:0)
12:21:48.391 | You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
12:21:48.391 | > #!/usr/bin/env bash
12:21:48.391 | \| #
12:21:48.392 | \| # create-clog.sh <version> <release-date>
```

scripts/template_changelog.md
vapurrmaid: Worth investigating - CH story would be useful as that's the responsibility of `m`, not anything in `cdr/docs` AFAICT

PR #202, opened 2021-03-25
clarify supported git providers

ericpaulsen:
adding this clarification, as it's come in up a [slack thread](https://codercom.slack.com/archives/C014JH42DBJ/p1616584356048200) that users can utilize any git provider, as long as they manage their own SSH keys. users are only limited to the three listed git providers if they expect to configure a git integration via OAuth.

### DIFF
--- a/admin/git.md
+++ b/admin/git.md
@@ -17,6 +17,9 @@ Currently, Coder supports the following providers:
   rel="noreferrer noopener">doesn't support</a> managing SSH keys for users via
   OAuth)
 
+**Note:** Utilizing one of the above git providers via OAuth integration isn't required.
+You can use any git provider of choice given you manage your own SSH keys. 
+
 ![Configure Git Integration](../assets/git-admin.png)
 
 ## Configuring OAuth
--- a/admin/git.md
+++ b/admin/git.md
@@ -8,7 +8,8 @@ their accounts with the Git repository service of choice.
 
 ## Support
 
-Currently, Coder integrates with the following git providers:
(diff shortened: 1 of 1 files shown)


admin/git.md
```diff
@@ -8,7 +8,7 @@ their accounts with the Git repository service of choice.
 
 ## Support
 
-Currently, Coder supports the following providers:
+Currently, Coder integrates with the following git providers:
```
jawnsy, line 11:
```suggestion
Coder integrates with the following service providers for authentication (use as an identity provider) as well as [user key management](https://coder.com/docs/environments/preferences#linked-accounts):
```

admin/git.md
```diff
@@ -17,6 +17,9 @@ Currently, Coder supports the following providers:
   rel="noreferrer noopener">doesn't support</a> managing SSH keys for users via
   OAuth)
 
+> You can use any git provider of choice. Configuring a git integration provides
```
jawnsy, line 20:
```suggestion
These integrations are optional. Linking your Coder account with your service provider provides support for key management features, but is not mandatory. We expect that Visual Studio Code and the `git` command-line tool will interoperate with any compliant hosting software or service, but Coder does not test these regularly and cannot provide recommendations or support.
```

jawnsy (commented):
LGTM, but @khorne3 will need to take a look

khorne3 (approved):

khorne3:
Thank you for your help @jawnsy!

ericpaulsen:
Thank you both!

PR #208, opened 2021-03-26
chore: Automate adding clog to manifest

vapurrmaid:
## Context

In c886756b2783d1289eaaf96b950c05d02585e26b a clog generation script was added in order to help automate changelog generation. However, generated clogs still needed to be manually added to `manifest.json`. At the time, use of `jq` was experimented with in the `create-clog.sh` script, but was found to be not worth the effort.

## Summary

Adds a NodeJS script written in TS for adding changelog entries to `manifest.json`.

## Details

The `manifest.json` file was formatted as part of this diff because the `add_clog_to_manifest.ts` script uses `JSON.stringify` to serialize changes to JSON. The resulting format differed from our prior format, but still complies with prettier.

## Example

```console
$ ./scripts/create-clog.sh --version=1.18.2 --release-date=$(date +%D)
Creating template using version: 1.18.2 and release date: 03/26/21
Calling add_clog_to_manifest.ts to update manifest.json
appending manifest.json with clog version:  1.18.2
done
tip: run yarn fmt:fix to prettify manifest.json
Done
```

```bash
git diff
```

```diff
diff --git a/manifest.json b/manifest.json
index 12cfaf2..27b98c2 100644
--- a/manifest.json
+++ b/manifest.json
@@ -309,6 +309,9 @@
     {
       "path": "./changelog/index.md",
       "children": [
+        {
+          "path": "./changelog/1.18.2.md"
+        },
         {
           "path": "./changelog/1.17.1.md"
         },
```

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -3,9 +3,15 @@
     {
       "path": "./index.md",
       "children": [
-        { "path": "about.md" },
-        { "path": "comparison.md" },
-        { "path": "feedback.md" }
+        {
+          "path": "about.md"
+        },
+        {
+          "path": "comparison.md"
+        },
+        {
+          "path": "feedback.md"
+        }
       ]
     },
     {
@@ -34,43 +40,87 @@
     {
       "path": "./images/index.md",
       "children": [
-        { "path": "./images/importing.md" },
-        { "path": "./images/structure.md" },
-        { "path": "./images/tags.md" },
-        { "path": "./images/configure.md" },
-        { "path": "./images/ssl-certificates.md" },
-        { "path": "./images/deprecating.md" },
-        { "path": "./images/embed-button.md" }
+        {
+          "path": "./images/importing.md"
+        },
+        {
+          "path": "./images/structure.md"
+        },
+        {
+          "path": "./images/tags.md"
+        },
+        {
+          "path": "./images/configure.md"
+        },
+        {
+          "path": "./images/ssl-certificates.md"
+        },
+        {
+          "path": "./images/deprecating.md"
+        },
+        {
+          "path": "./images/embed-button.md"
+        }
       ]
     },
     {
       "path": "./cli/index.md",
       "children": [
-        { "path": "./cli/installation.md" },
-        { "path": "./cli/remote-terminal.md" },
-        { "path": "./cli/file-sync.md" },
-        { "path": "./cli/managing-envs.md" }
+        {
+          "path": "./cli/installation.md"
+        },
+        {
+          "path": "./cli/remote-terminal.md"
+        },
+        {
+          "path": "./cli/file-sync.md"
+        },
+        {
+          "path": "./cli/managing-envs.md"
+        }
       ]
     },
     {
       "path": "./setup/index.md",
       "children": [
-        { "path": "./setup/architecture.md" },
-        { "path": "./setup/requirements.md" },
+        {
+          "path": "./setup/architecture.md"
+        },
+        {
+          "path": "./setup/requirements.md"
+        },
         {
           "path": "./setup/kubernetes/index.md",
           "children": [
-            { "path": "./setup/kubernetes/local-preview.md" },
-            { "path": "./setup/kubernetes/aws.md" },
-            { "path": "./setup/kubernetes/azure.md" },
-            { "path": "./setup/kubernetes/google.md" }
+            {
+              "path": "./setup/kubernetes/local-preview.md"
+            },
+            {
+              "path": "./setup/kubernetes/aws.md"
+            },
+            {
+              "path": "./setup/kubernetes/azure.md"
+            },
+            {
+              "path": "./setup/kubernetes/google.md"
+            }
           ]
         },
-        { "path": "./setup/installation.md" },
-        { "path": "./setup/configuration.md" },
-        { "path": "./setup/licensing.md" },
-        { "path": "./setup/air-gapped.md" },
-        { "path": "./setup/updating.md" }
+        {
+          "path": "./setup/installation.md"
+        },
+        {
+          "path": "./setup/configuration.md"
+        },
+        {
+          "path": "./setup/licensing.md"
+        },
+        {
+          "path": "./setup/air-gapped.md"
+        },
+        {
+          "path": "./setup/updating.md"
+        }
       ]
     },
     {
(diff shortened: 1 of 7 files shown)


vapurrmaid:
Looks like the build workflow based off of `m` needs to be addressed:

> ```console
> Failed to compile.
>   
>   ./docs/scripts/add_clog_to_manifest.ts:7:23
>   Type error: Could not find a declaration file for module 'minimist'. '/vercel/workpath0/product/coder.com/site/node_modules/minimist/index.js' implicitly has an 'any' type.
>     Try `npm i --save-dev @types/minimist` if it exists or add a new declaration (.d.ts) file containing `declare module 'minimist';`
>   
>      5 |  */
>      6 | import { writeFileSync } from "fs";
>   >  7 | import parseArgs from "minimist";
>        |                       ^
>      8 | import { resolve } from "path";
>      9 | import semverCompare from "semver-compare";
>     10 | import manifestJSON from "../manifest.json";
>  ```

The `m` build shouldn't be looking into dev helper scripts defined in this repository.

Will investigate when have bandwidth.

manifest.json
```diff
@@ -3,68 +3,142 @@
     {
       "path": "./index.md",
       "children": [
-        { "path": "about.md" },
-        { "path": "comparison.md" },
-        { "path": "feedback.md" }
+        {
```
jawnsy, line 6: formatting changes look good to me, but we should check with @khorne3 since she has to look at this stuff more than I do :joy:

scripts/add_clog_to_manifest.ts
```diff
@@ -0,0 +1,96 @@
+/**
+ * @file add_clog_to_manifest.ts
+ * @description Adds a changelog entry to manifest.json
+ * @example add_clog_to_manifest.ts --version=1.18.3
+ */
+import { writeFileSync } from "fs";
```
jawnsy, line 6: this is a nitpick but the style we use in product/coder is to skip trailing semicolons, can we copy over our prettierrc so that formatting is consistent? (assuming it actually fixes that)

no worries if not, this is perfectly cromulent as-is

jawnsy (approved):

scripts/add_clog_to_manifest.ts
vapurrmaid: Sure thing, we can update 1 line in `.prettierrc` and run `fmt:fix` :D

manifest.json
vapurrmaid: :D

vapurrmaid:
The CI failures for running `next build` should disappear after some time - I imagine the `m/` version pulled in is cached/takes time to pull in changes or the node_modules install step is cached on CI.

package.json
```diff
@@ -1,24 +1,31 @@
...
   "scripts": {
-    "fmt": "prettier --check '**/*.{json,md,yaml,yml}'",
-    "fmt:fix": "prettier --write '**/*.{json,md,yaml,yml}'",
+    "fmt": "prettier --check '**/*.{json,md,ts,yaml,yml}'",
+    "fmt:fix": "prettier --write '**/*.{json,md,ts,yaml,yml}'",
     "lint": "markdownlint '**/*.md'",
     "lint:fix": "markdownlint --fix '**/*.md'"
   },
   "devDependencies": {
+    "@types/minimist": "^1.2.1",
```
jawnsy, line 9: @vapurrmaid Opening as a reply here so that we can have a threaded discussion

>The CI failures for running next build should disappear after some time - I imagine the m/ version pulled in is cached/takes time to pull in changes or the node_modules install step is cached on CI.

If there's something wrong with CI and caching, we should fix so that we can trust it. Otherwise we can end up in a situation where merging this might cause coder.com or dashboard builds to start failing?

I think the problem might be that the monorepo root package.json needs these dependencies added, or perhaps the one in `product/coder.com/site`? The preview CI clones the monorepo and then updates the docs repo, but runs the build from the coder.com site root, so my guess is that one needs to be updated to make nextjs happy

jawnsy: This build is kinda janky because we can't actually build docs independently (they are built as a subpage of the full site)... so I think we need to add these deps to https://github.com/cdr/m/blob/bdd808f0bd10252503b8cee86b71bf329d579f85/product/coder.com/site/package.json#L11 as a separate PR first, get that merged into m, then the previews should hopefully work and will be safe to merge?

vapurrmaid: @jawnsy I appreciate the thread and confidence in CI - I don't oversee what/how node_modules are cached, but am confident it's related to that due to the below:

```console
$ cd product/coder.com/site/docs/
$ git checkout origin/vapurrmaid/manifest
Previous HEAD position was 6cd3673 Add 1.17.1 changelog (#206)
HEAD is now at 820897d chore: Do not use semis
$ cd ..
$ yarn build
```

Which succeeds

> Otherwise we can end up in a situation where merging this might cause coder.com or dashboard builds to start failing?

This is only true if we bump this submodule in `m/` which will also need to pass CI

jawnsy: Did you also go up to coder.com and run the yarn build from there?

jawnsy: Nevermind, misread your session. Well alright!

vapurrmaid: I'll re-run the action. If it fails, another possible course of action is closing this PR and opening a new one

jawnsy: This seems to repro for me using your branch. Here's what I did:

1. Check out the monorepo
2. Go to product/coder/docs
3. Check out this branch
4. Go back up to coder
5. Run yarn, then yarn build

```
[coder@jawnsy-m site]$ yarn
yarn install v1.22.10
[1/4] Resolving packages...
success Already up-to-date.
Done in 0.43s.
[coder@jawnsy-m site]$ yarn build
yarn run v1.22.10
$ next build
warn  - React 17.0.1 or newer will be required to leverage all of the upcoming features in Next.js 11. Read more: https://err.sh/next.js/react-version
Loaded env from /home/coder/projects/m/product/coder.com/site/.env
Failed to compile.

./docs/scripts/add_clog_to_manifest.ts:7:23
Type error: Could not find a declaration file for module 'minimist'. '/home/coder/projects/m/product/coder.com/site/node_modules/minimist/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/minimist` if it exists or add a new declaration (.d.ts) file containing `declare module 'minimist';`

   5 |  */
   6 | import { writeFileSync } from "fs";
>  7 | import parseArgs from "minimist";
     |                       ^
   8 | import { resolve } from "path";
   9 | import semverCompare from "semver-compare";
  10 | import manifestJSON from "../manifest.json";
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
[coder@jawnsy-m site]$ 
```

vapurrmaid: > Go to product/coder/docs

Should be `coder.com`

> Go back up to coder

Either need to run `yarn` in `coder.com/site/docs` or in root of `m/`. Running it anywhere else won't install these node_modules

jawnsy: we have a fascinating build system, wow. thanks for the explanation. hmm. I don't have any objection to just merging this and seeing if it works down the line in the submodule build, in case it is indeed just a caching issue.

well I think the preview job clones m and runs stuff from there, if I'm interpreting this correctly: https://github.com/cdr/docs/blob/6ec7c2313588fdfafea6a67121c60f98f8c0efb8/.github/workflows/preview.yaml#L27

The vercel deploy step seems to be happening from the root of the monorepo, which also kinda confuses me (I would've thought that we were deploying from product/coder.com/site instead)

vapurrmaid: If any grumpiness from me is coming out, it's because I'd strongly prefer less mono-repoing and less submodularizing and more package.json dependency-ing lol

vapurrmaid: Interesting RE vercel...that seems odd to me as well :confused:

jawnsy: lol it is totally understandable given this particular situation (and I am sure many more you have encountered in the past)... I'm just wondering what the problem is and how we fix our build to avoid it. I'm also open to your original theory that this is a caching issue, there's just some risk we need to be aware of there, that it could break m's docs submodule updates until it's fixed, though as you also point out, it won't affect the main site since our m/coder.com build should catch it

jawnsy: It is also entirely possible that I broke this when updating the build earlier

vapurrmaid: > I agree with that, yeah...
> 
> I didn't know about the post-install script magic until you pointed it out. I think Vercel is rebuilding everything from our source code, so I didn't think caching would play a part here?

vapurrmaid: ^ This is from the (now locked) #210 to avoid duplication

vapurrmaid: @jawnsy What do you think about us moving `cdr/docs` to a NPM package that onyl distributes a the packaged production files, rather than its development scripts? We could move off of using a submodule approach in `m/`, but we'd still get alerts from depend a bot. It also gives us easier roll-back-ability and versioning of docs

vapurrmaid: It has a further advantage of allowing clients/customers to download a zip of our documentation locked to a specific version

jawnsy: We're not currently using dependabot to track npm modules in m, but yeah that's certainly a possibility.

> The fact that coder.com/site is parsing a ts file in cdr/docs is weird to me.

Agree with you there, all of this is weird and kinda unexpected, ha.

Looking at the Vercel dashboard, it shows stuff checked out at the monorepo level, though I guess we have customized vercelignore to exclude irrelevant stuff like the product:

image

But this does mean that it should be running postinstall from the root, so I'm not sure why the dependencies are missing?

vapurrmaid: > it should be running postinstall from the root, so I'm not sure why the dependencies are missing?

My original theory is it would take some time to be able to pull the latest `m/` commit containing d318b93a1c414c988d0b3951b3f96aeaf65e6306

vapurrmaid: However, that seems debunked as `a6223b278c3a44713bd5ca4bdfad9a8df674fa1a` was pulled in on last run which is after https://github.com/cdr/m/commit/d318b93a1c414c988d0b3951b3f96aeaf65e6306

jawnsy: Well, every build is pulling the latest monorepo code at the start of the preview build: https://github.com/cdr/docs/blob/6ec7c2313588fdfafea6a67121c60f98f8c0efb8/.github/workflows/preview.yaml#L14-L21

khorne3 (approved):

PR #214, opened 2021-03-30
chore: use the same linter settings as the monorepo

jawnsy:
Use the same script names (e.g. yarn format:write) as we do in
the monorepo, so that it's easier to switch between projects.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "scripts": {
-    "fmt": "prettier --check '**/*.{json,md,ts,yaml,yml}'",
-    "fmt:fix": "prettier --write '**/*.{json,md,ts,yaml,yml}'",
+    "prepare": "husky install",
+    "format:check": "prettier --check '**/*.{css,html,js,json,jsx,ts,tsx,yaml,yml}'",
+    "format:write": "prettier --write '**/*.{css,html,js,json,jsx,ts,tsx,yaml,yml}'",
     "lint": "markdownlint '**/*.md'",
-    "lint:fix": "markdownlint --fix '**/*.md'",
-    "prepare": "husky install"
+    "lint:fix": "markdownlint --fix '**/*.md'"
   },
   "devDependencies": {
     "@types/minimist": "^1.2.1",
@@ -21,7 +21,9 @@
     "typescript": "^4.2.3"
   },
   "lint-staged": {
-    "*.{json,ts,yaml,yml}": "prettier --write",
+    "*.{css,html,js,json,jsx,ts,tsx,yaml,yml}": [
+      "prettier --write"
+    ],
     "*.md": [
       "markdownlint --fix",
       "prettier --write"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -18,10 +18,11 @@
     }
   },
   "lint-staged": {
-    "*.{json,yaml,yml}": "prettier --write",
-    "*.md": [
-      "markdownlint --fix",
+    "*.{css,html,json,md,yaml,yml}": [
```
vapurrmaid, line 21: add `,ts`

jawnsy: I copied/merged these from the monorepo and would like to be consistent between them, so will update this to add js,ts,tsx and reflect the same extensions over in the monorepo root. Thanks for the review!

jawnsy: oh question though, should we move the markdownlint step before the formatting step? and maybe it's better to format it like we do with the monorepo, e.g.

```markdown
  "lint-staged": {
    "*.{css,html,js,json,jsx,ts,tsx,yaml,yml}": [
      "prettier --write"
    ],
    "*.md": [
      "markdownlint --fix",
      "prettier --write"
    ]
  }
```

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "scripts": {
-    "fmt": "prettier --check '**/*.{json,md,yaml,yml}'",
-    "fmt:fix": "prettier --write '**/*.{json,md,yaml,yml}'",
+    "format:check": "prettier --check '**/*.{css,html,js,json,md,scss,ts,tsx}'",
```
jawnsy, line 3: lmao, I just noticed that these formatters are different from the ones below in lint-staged :man_facepalming: I don't suppose there's some way to invoke one from the other or vice versa?

vapurrmaid: Yea, having to remember to update both spots is not fun. I can't think of a way off the top of my head unfortunately :(

vapurrmaid (approved):

package.json
vapurrmaid: If you find a way...that would be great. Maybe a raindy friday search topic!

package.json
jawnsy: yea, makes sense. I was trying to combine them to avoid repeating things but the existing approach is clearer

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "scripts": {
-    "fmt": "prettier --check '**/*.{json,md,yaml,yml}'",
-    "fmt:fix": "prettier --write '**/*.{json,md,yaml,yml}'",
+    "format:check": "prettier --check '**/*.{css,html,js,json,jsx,ts,tsx,yaml,yml}'",
+    "format:write": "prettier --write '**/*.{css,html,js,json,jsx,ts,tsx,yaml,yml}'",
     "lint": "markdownlint '**/*.md'",
     "lint:fix": "markdownlint --fix '**/*.md'",
     "prepare": "husky install"
```
jawnsy, line 7: going to move prepare up to the top (and keep all the lifecycle hooks above the other scripts)

PR #217, opened 2021-03-30
fix configure step formatting

ericpaulsen:
I'm not positive if this is a "fix" or if the prior formatting was intended. I just made the change to make sure all of the steps match.

Let me know. Thanks!


vapurrmaid:
The error is being observed on `README.md`

vapurrmaid (approved):
@ericpaulsen should be good now

khorne3 (approved):

PR #222, opened 2021-03-30
Document network policy requirement [ch9751]

ericpaulsen:
Adding CNI requirement to Setup > Requirements, as customers have deployed Coder without network policies enforced, leading to surprises about security holes in their clusters.

Open to edits & feedback.

[Clubhouse story](https://app.clubhouse.io/coder/story/9751/document-network-policy-requirement)

### DIFF
--- a/setup/requirements.md
+++ b/setup/requirements.md
@@ -45,17 +45,28 @@ launches the Remote IDE in a pop-up window.
 
 ## Network Policies
 
-Coder uses [Kubernetes Network Policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) to enforce network segmentation and tenant isolation within your cluster. Coder's network isolation policy blocks all ingress traffic to workspaces, with the exception of traffic from the control plane, thereby ensuring that all traffic is auditable. The control plane does not specify egress rules, which allows outbound traffic by default, while allowing you to enforce a more specific network policy if you wish.
-[Container Network Interface (CNI)](https://github.com/containernetworking/cni#what-is-cni)
-plugins to implement network segmentation and tenant isolation in the Kubernetes
-cluster. This enforces network boundaries between pods, preventing users from
+Coder uses
+[Kubernetes NetworkPolicies](https://kubernetes.io/docs/concepts/services-networking/network-policies/)
+to enforce network segmentation and tenant isolation within your cluster.
+
+Coder's network isolation policy blocks all ingress traffic to workspaces except
+traffic from the control plane (this ensures that you can audit all traffic).
+However, the control plane does not specify egress rules; by default, it allows
+outbound traffic. However, you can still enforce a more specific network policy.
+
+[Container network interface (CNI)](https://github.com/containernetworking/cni#what-is-cni)
+plugins implement network segmentation and tenant isolation in the Kubernetes
+cluster. They enforce network boundaries between pods, preventing users from
 accessing other workspaces.
 
-If your Container Network Interface (CNI) plugin does not support NetworkPolicy enforcement, traffic between workspaces and other containerized workloads within the same cluster will be permitted to communicate without restriction. Consider testing your container networking after installation to ensure that the behavior is as you expect.
-be able to communicate with each other without restriction.
+If your container network interface (CNI) plugin does not support NetworkPolicy
+enforcement, traffic between workspaces, and other containerized workloads
+within the same cluster will be permitted to communicate without restriction.
+Consider testing your container networking _after_ installing Coder to ensure
+that the behavior is as expected.
 
-See more:
-[Kubernetes Network Policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/)
+> If you're not sure which CNI plugin, we suggest
+> [Calico](https://docs.projectcalico.org/getting-started/kubernetes/quickstart).
 
 ## Licenses
 
--- a/setup/requirements.md
+++ b/setup/requirements.md
@@ -43,11 +43,13 @@ currently require the following versions _or newer_:
 If you're using [Remote IDEs](../environments/editors.md), allow pop-ups; Coder
 launches the Remote IDE in a pop-up window.
 
-## Container Network Interface (CNI)
+## Network Policies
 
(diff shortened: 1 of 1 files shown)


setup/requirements.md
```diff
@@ -43,6 +43,20 @@ currently require the following versions _or newer_:
...
+
+Coder uses
+[Container Network Interface (CNI)](https://github.com/containernetworking/cni#what-is-cni)
+plugins to implement network segmentation and tenant isolation in the Kubernetes
+cluster. This enforces network boundaries between pods, preventing users from
+accessing other workspaces.
+
+If there is no CNI plugin configured to enforce such policies, workspaces will
+be able to communicate with each other without restriction.
+
```
khorne3, line 56: I'm assuming there's an actionable step here; if so, we could make it clearer to the reader and say, "If you don't configure a CNI plugin to enforce such policies, workspaces will..." and then maybe some info regarding how they might do this (maybe we could tell people that we use Calico and here's how?)

When we say workspaces, we actually mean workspaces (as in environments), right?

ericpaulsen: > When we say workspaces, we actually mean workspaces (as in environments), right?

Correct. And yes, the next actionable step would be to enable a CNI plugin when creating the cluster. Both Azure & GCP have Calico built in, but it requires enabling in the create cluster commands. Enabling Calico in AWS requires some separate commands after the cluster is created.

I'm thinking we say something to the effect that "you'll need to include a CNI plugin compatible with your Kubernetes provider when creating your cluster"

setup/requirements.md
```diff
@@ -43,6 +43,20 @@ currently require the following versions _or newer_:
...
 
+## Network Policies
+
+Coder uses
+[Container Network Interface (CNI)](https://github.com/containernetworking/cni#what-is-cni)
+plugins to implement network segmentation and tenant isolation in the Kubernetes
+cluster. This enforces network boundaries between pods, preventing users from
+accessing other workspaces.
+
+If there is no CNI plugin configured to enforce such policies, workspaces will
```
jawnsy, line 54: I think that CNI plugins are required for Kubernetes to work (these interfaces are just gRPC interfaces that separate what Kubernetes is trying to do from the implementation), but not all CNI plugins support enforcing network policies. I think this should be in an aside or with a warning symbol because this is pretty important for production installs

```suggestion
If your Container Network Interface (CNI) plugin does not support NetworkPolicy enforcement, traffic between workspaces and other containerized workloads within the same cluster will be permitted to communicate without restriction. Consider testing your container networking after installation to ensure that the behavior is as you expect.
```

setup/requirements.md
jawnsy: We may want to specify a recommendation based on what we test (or state that we test with a specific CNI provider, whatever that is), for the people that are unfamiliar with Kubernetes and don't know which CNI provider to choose.

setup/requirements.md
```diff
@@ -43,6 +43,20 @@ currently require the following versions _or newer_:
...
+be able to communicate with each other without restriction.
+
+See more:
```
jawnsy, line 57: **suggestion**: consider collating these links at the bottom of the page in a "See also" section, or turn this into a "Learn more about Network Policies" link?

setup/requirements.md
ericpaulsen: That works with me - I also think we should take @jawnsy's edits into account when saying "Be sure to include..." by making clear that the customer should use a CNI that supports NetworkPolicies.

setup/requirements.md
ericpaulsen: Good distinction here - I was not aware of this previously. We should include this @khorne3.

setup/requirements.md
ericpaulsen: Good with me.

ericpaulsen: > We may want to specify a recommendation based on what we test (or state that we test with a specific CNI provider, whatever that is), for the people that are unfamiliar with Kubernetes and don't know which CNI provider to choose.

Agreed, I wasn't sure if we wanted to get into recommending CNI providers, but since Calico it is already built into GCP & Azure, I'd say it's accredited enough to recommend here.

jawnsy: Yeah, I'm not sure about the idea of recommending things either - the only reason I mentioned it was for the use case of "I want to try Coder but I don't know much about Kubernetes"

We could say "If you're not sure, we suggest choosing Calico" or something to that effect? Any CNI provider should work and it's not a recommendation per se (because we're not saying to choose Calico over Cilium or something else, if your organization standardizes on those).

My intent would just be to nudge users toward whatever we test with, because it's more likely to work

ericpaulsen: Agree - let's go with the "if you're not sure" verbiage, as it isn't a blatant recommendation.

setup/requirements.md
jawnsy: Actually, on second thought, maybe we could be more vague and say "If your cluster networknig configuration does not enforce Network Policies ..."

Also, I suggest linking to this page instead of the CNI spec open source repo (that one is more useful for implementers of CNI plugins): https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/

khorne3:
Should we simplify this to `Enforce network policies using the CNI plugin of your choice` then link to a Guides page with the explanation that we've been working on?

ericpaulsen:
> Should we simplify this to `Enforce network policies using the CNI plugin of your choice` then link to a Guides page with the explanation that we've been working on?

While this would definitely simplify things, I think keeping the explanation & context in `Requirements` sends a clear message that this topic requires consideration up front before deploying a cluster & installing Coder.

Unlikely that we'll get 100% of users to navigate away from `Requirements` to read the necessary NetworkPolicy info.

Thoughts?

khorne3:
@ericpaulsen That's a good point. I also was confused for a moment; this is on the requirements page (where we expect people to read) not the install page (where I feel like this much explanation would be jarring).

khorne3 (approved):

PR #223, opened 2021-03-30
chore: Add post-checkout hook

vapurrmaid:
LMK if you think we should add the `yarn.lock` check. I didn't feel compelled to because this project is small enough, but happy to modify if we prefer it


jawnsy (approved):
Nice

PR #232, opened 2021-04-01
chore: clarify version support policy

jawnsy:
Recommend that users track supported Kubernetes versions, and remove
the reference to version 1.13.

[ch13998]

### DIFF
--- a/setup/kubernetes/index.md
+++ b/setup/kubernetes/index.md
@@ -3,24 +3,28 @@ title: Kubernetes
 description: Learn how to set up a Kubernetes cluster for your Coder deployment.
 ---
 
-You can deploy Coder into any [compatible Kubernetes cluster]. Coder follows the
-[Kubernetes upstream version support policy], and the latest stable release
-version of Coder supports the previous two minor releases as well as the current
-release of Kubernetes at time of publication. Coder may run successfully with
-older versions of Kubernetes, however, we strongly recommend running one of the
-currently-supported versions, in order to receive applicable fixes, including
-security fixes, from the Kubernetes project maintainers.
-
-Coder continuously removes usage of deprecated Kubernetes API versions once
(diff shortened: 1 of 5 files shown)


setup/kubernetes/index.md
```diff
@@ -1,13 +1,21 @@
 ---
 title: Kubernetes
-description: Learn how to set up a K8 cluster for your Coder deployment.
+description: Learn how to set up a Kubernetes cluster for your Coder deployment.
 ---
 
-You can install and deploy Coder onto any Kubernetes cluster (version 1.13.7+)
-that meets the [requirements](../requirements.md).
+You can deploy Coder into any [compatible Kubernetes cluster]. We recommend
+following the [Kubernetes upstream version support policy] in order to receive
```
jawnsy, line 7: I wrote kind of a vague statement here instead of a specific version, but I think we should explicitly _require_ that people use a supported version of Kubernetes, so that we can rely on Kubernetes features available in 1.18 (currently the oldest version in support).

khorne3: Maybe we tie it in to K8 support -- if they do, we do (so we'd support the three minor releases K8 does); then we could write a generic statement and link to this [doc](https://kubernetes.io/docs/setup/release/version-skew-policy/)

khorne3:
@jawnsy Do you mind if I include in this PR removals of old version references and such?

jawnsy:
@khorne3 That would be wonderful, thank you! I've rebased so that the branch is up-to-date. I still need to take another pass at this given our new version support policy.

I'd suggest that we remove version numbers elsewhere in the documentation, and use a term like "compatible Kubernetes cluster" linking to this page, so that we don't have to update a bunch of versions later

khorne3 (approved):

jawnsy:
Thanks for the edits!

PR #235, opened 2021-04-01
update GCP node type in create cluster command

ericpaulsen:
updating node type to `n1-highmem-4` as it better supports the use of CVMs and removes scale up issues during env builds. uses the same amount of vCPUs as the prior `n1-standard-4` just with added memory. 

we've seen customers deploy via our happy path and realize they need a bigger node after the fact. hopefully this will avoid that issue.


khorne3 (approved):

bpmct (approved):
👏🏼

PR #236, opened 2021-04-02
clarify GKE cluster options

ericpaulsen:
adding these edits to make it clearer to the reader that there are two options when deploying a GKE cluster.

### DIFF
--- a/setup/kubernetes/google.md
+++ b/setup/kubernetes/google.md
@@ -14,9 +14,14 @@ machine and configured to interact with your Google Cloud Platform account.
 
 ## Set Up the GKE Cluster
 
-The following will spin up a Kubernetes cluster using the `gcloud` command (be
-sure to replace the parameters (specifically `PROJECT_ID`, `NEW_CLUSTER_NAME`,
-and `ZONE`) as needed to reflect the needs of your environment).
+The following commands will spin up a Kubernetes cluster using the `gcloud`
+command (be sure to replace the parameters (specifically `PROJECT_ID`,
+`NEW_CLUSTER_NAME`, and `ZONE`) as needed to reflect the needs of your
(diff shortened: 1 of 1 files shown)


setup/kubernetes/google.md
```diff
@@ -14,9 +14,14 @@ machine and configured to interact with your Google Cloud Platform account.
 
 ## Set Up the GKE Cluster
 
-The following will spin up a Kubernetes cluster using the `gcloud` command (be
-sure to replace the parameters (specifically `PROJECT_ID`, `NEW_CLUSTER_NAME`,
-and `ZONE`) as needed to reflect the needs of your environment).
+The following commands will spin up a Kubernetes cluster using the `gcloud`
```
jawnsy, line 17: general suggestions with this document:

1. Mention the settings we need in a table at the top, users can also create clusters manually in the user interface rather than using the gcloud command, and it's probably easier for users new to Kubernetes. We shouldn't provide instructions for doing the graphical-driven install, since the UI is likely to change, but it's worth mentioning as an option for those that prefer it.
2. Avoid nested parentheticals
3. Add a separate planning step somewhere (maybe under Requirements and link to it from here) to describe our requirements - for example for CVMs we need Ubuntu; otherwise either COS or Ubuntu are acceptable. As a general approach though we test most with Ubuntu, so I think that's what we should recommend. We use Ubuntu with Docker internally but I think we should be using Ubuntu with containerd and I think that should work fine
4. Instead of describing what the options do in the intro, we can move the description as a header, for example:
 
### Option 1: Full support of Coder features

### Option 2: Minimum requirements

(I also suggest putting the recommended option first, since that's the one a user would be more likely to pick)

khorne3 (approved):

PR #237, opened 2021-04-03
Update air-gapped / offline installation

mterhar:
After going through the process, many items seem relevant that were omitted. There's an argument to be made that the majority of what I'm suggesting here is out of scope for docs and may do better in Guides. I'm happy to have that discussion and move things around, but I wanted to get the draft ironed out while it was fresh.

[ch10080]


setup/air-gapped.md
```diff
@@ -60,6 +71,22 @@ platform images are hosted in Coder's Docker Hub repo.
    docker pull coderenvs/coder-service:<version>
    ```
 
+   Additional images may be needed to configure and run workspaces:
```
khorne3, line 74: This line makes it sound like the following images are optional. @mterhar I *think* all of these are required so that they have all Coder features, right?

If these images are required, then we should tell users to install them; if they're not, we should explain when things are required

ericpaulsen: Outside of the ingress controller, all of the listed images are optional, as customers can build their own node & IntelliJ images (for example) and have the option to hook into the OpenVSX extension marketplace. 

That said, while bringing in an _nginx_ ingress controller is optional, I believe we require an ingress in order to access Coder's external IP. So we should clarify that if you are not using an nginx ingress, you'll be responsible for providing your own ingress controller.

setup/air-gapped.md
```diff
@@ -87,9 +114,171 @@ platform images are hosted in Coder's Docker Hub repo.
...
+For a self-signed certificate to be passed into the Coder images a secret needs
+to be created and then referenced in the chart. The secret is a created from a
+directory with a single file in it. The directory name is irrelevant and the
+filename becomes the `key` if the following command is used.
+
+```bash
+kubectl -n coder create secret generic local-registry-cert --from-file=/certs
+```
+
+If the `-out` argument on the OpenSSL command to generate the certificates was
```
khorne3, line 214: We used -keyout, but I'm assuming this still applies?

khorne3:
FYI I'm in the middle of splitting this into two docs; the current air-gap doc will be mostly the same with the infra stuff on its own.

setup/air-gapped/infrastructure.md
```diff
@@ -0,0 +1,159 @@
...
+
+> For the volume mounted at `/var/lib/registry` make sure it can store 10+ GB
+> for just Coder images.
+
+## Configuring the Kubernetes Node
+
+Before the Kubernetes node can accept your certificate, you'll need to mark your
+`registry.crt` file as trusted. The specific locations where you need to have
+this file depends on your Linux distribution and the container runtime:
+
```
khorne3, line 60: I thought this section was a bit confusing; can you expand a bit more on what the end-user needs to consider so the cert's trusted?

khorne3:
@mterhar @ericpaulsen (Eric, I'm tagging you since I think you have experience with this?) -- I split this into to docs and moved some stuff around based on where I thought things made sense. I also had some questions, so hopefully, you could provide additional guidance?

setup/air-gapped/index.md
```diff
@@ -0,0 +1,171 @@
...
+1. Pull the images for the Coder platform from the following Docker Hub
+   locations:
+
+   [coder-service](https://hub.docker.com/r/coderenvs/coder-service)
+
+   [envbox](https://hub.docker.com/r/coderenvs/envbox)
+
+   [envbuilder](https://hub.docker.com/r/coderenvs/envbuilder)
+
+   [timescale](https://hub.docker.com/r/coderenvs/timescale)
```
ericpaulsen: We really only use timescale for evaluation purposes, and in our helm chart, we recommend that customers bring in their own Postgres, and keep it external to the cluster. 

I think we should clarify this up-front, rather than instructing customers to pull in the timescale image (it also has a bevy of critical vulnerabilities).

setup/air-gapped/infrastructure.md
ericpaulsen: I'm not experienced with this piece just yet, so I will defer to @mterhar.

manifest.json
```diff
@@ -63,7 +63,10 @@
         { "path": "./setup/installation.md" },
         { "path": "./setup/configuration.md" },
         { "path": "./setup/licensing.md" },
-        { "path": "./setup/air-gapped.md" },
+        {
+          "path": "./setup/air-gapped/index.md",
+          "children": [{ "path": "./setup/air-gapped/infrastrucutre.md" }]
```
jawnsy, line 68: seems to be a typo in the filename, infrastructure?

setup/air-gapped/infrastructure.md
mterhar: The multiple locations in the file below are all the places that I needed to set for an ubuntu base image to be okay with the self-signed cert. Maybe just referring to that list as a partial set that may require additions depending on the OS and kubernetes distribution.

setup/air-gapped/infrastructure.md
```diff
@@ -0,0 +1,160 @@
...
+    -v /var/lib/docker/registry:/var/lib/registry \
+    registry:2
+```
+
+> For the volume mounted at `/var/lib/registry` make sure it can store 10+ GB
+> for just Coder images.
+
+## Configuring the Kubernetes Node
+
+Before the Kubernetes node can accept your certificate, you'll need to mark your
```
mterhar, line 57:
```suggestion
Before the Kubernetes node can accept run local images, it needs to consider the new
```

setup/air-gapped/infrastructure.md
```diff
@@ -0,0 +1,160 @@
...
+update-ca-certificates
+cat <<EOT >> /etc/containerd/config.toml
+[plugins."io.containerd.grpc.v1.cri".registry.configs."$REGISTRY_DOMAIN_NAME".tls]
+  insecure_skip_verify = true
+EOT
+systemctl restart containerd
+```
+
+Because you'll need to run the steps described in this section on all nodes that
+will be scheduling Coder images, we recommend that you either:
```
mterhar, line 82:
```suggestion
Because the steps described in this section must be run on all nodes which
will be scheduling Coder images, it is best to:
```

setup/air-gapped/infrastructure.md
```diff
@@ -0,0 +1,160 @@
...
+1. Include these steps in the image itself
+1. Run an init script including these instructions whenever you add a new node
+   to your cluster
+
+## Adding certificate secrets to the Helm chart
+
+Coder validates images and pulls tags using API calls, so issues with your
+certificates may prevent you from adding images. If you see such issues and you
+have a certificate authority in your network, you may need to add the root
+certificate.
```
mterhar, line 93:
```suggestion
Coder validates images and pulls tags using direct REST API calls to the registry.
Other internal services (OICD, Git providers, etc) which use HTTPS APIs require
the Coder container to trust the certificate. The Coder helm chart can be used to
add a root CA certificate to the Coder service images.
```

setup/air-gapped/infrastructure.md
```diff
@@ -0,0 +1,160 @@
...
+1. Create a secret
+1. Reference the secret in your Coder Helm chart
+
+To create a secret, run:
+
+```console
+kubectl -n coder create secret generic local-registry-cert --from-file=/certs
+```
+
+When using the above command, you're creating the secret from a directory with a
```
mterhar, line 106:
```suggestion
When using the above command, `kubectl` creates the secret from a directory with a
```

setup/air-gapped/infrastructure.md
```diff
@@ -0,0 +1,160 @@
...
+single file. The directory name doesn't matter, but the filename becomes the
+secret **key**.
+
+> If you changed the `-out` argument on the OpenSSL command used to generate the
+> certificates, or if you moved the certificates, make sure that you adjust the
+> path included with `--from-file=`.
+
+To verify your secret:
```
mterhar, line 114:
```suggestion
To verify the new secret:
```

setup/air-gapped/infrastructure.md
```diff
@@ -0,0 +1,160 @@
...
+
+```console
+kubectl -n coder get secret local-registry-cert -o yaml
+```
+
+You'll need to add your secret to the Helm chart, and you can do this as part of
+your verification step. To do so, place the following snippet into a YAML file
+named `registry-cert-values.yml`:
```
mterhar, line 122:
```suggestion
Reference the newly secret from the Helm chart by adding the following
snippet into a YAML file named `registry-cert-values.yml`:
```

setup/air-gapped/infrastructure.md
```diff
@@ -0,0 +1,160 @@
...
+
+```console
+kubectl -n coder get secret local-registry-cert -o yaml -f registry-cert-values.yml
+```
+
+### Resolving the registry using the cluster's DNS or hostAliases
+
+Your Nodes must have their host file set to the `$REGISTRY_DOMAIN` and the
+static IP address of the local registry. For example, if the registry is on
+10.0.0.2, then add this to your Node configuration script:
```
mterhar, line 142:
```suggestion
Nodes must be able to resolve `$REGISTRY_DOMAIN` name to the static IP address
of the local registry. One way to do this without an external DNS server is to use the
node's hosts file. For example, if the registry is on 10.0.0.2, then add this to your Node
configuration script:
```

setup/air-gapped/infrastructure.md
```diff
@@ -0,0 +1,160 @@
...
+is a partial list to help you get started:
+
+```plaintext
+/usr/local/share/ca-certificates/registry.crt
+/etc/docker/certs.d/${REGISTRY_DOMAIN_NAME}/ca.crt
+/etc/ssl/certs/registry.crt
+/etc/pki/tls/registry.crt
+```
+
+If you're working with containerd, use the following to patch in certificates
```
mterhar, line 69:
```suggestion
If the cluster uses containerd, apply the following to patch in certificates
```

setup/air-gapped/infrastructure.md
```diff
@@ -0,0 +1,160 @@
...
+
+> For the volume mounted at `/var/lib/registry` make sure it can store 10+ GB
+> for just Coder images.
+
+## Configuring the Kubernetes Node
+
+Before the Kubernetes node can accept your certificate, you'll need to mark your
+`registry.crt` file as trusted. The specific locations where you need to have
+this file depends on your Linux distribution and the container runtime, but here
+is a partial list to help you get started:
```
mterhar, line 60:
```suggestion
`registry.crt` file as trusted. The specific locations and methods to store and
trust the certificate vary depending on the Linux distribution and the container 
runtime, but here is a partial list to start with:
```

setup/air-gapped/infrastructure.md
```diff
@@ -0,0 +1,160 @@
...
+> strongly recommend that you verify the snippets as well before using.
+
+## Creating the local registry and generating a self-signed certificate
+
+You will need to create a local registry to store your Coder images and
+self-signed certificates (you can use self-signed certificates in any
+environment, but they're required for air-gapped deployments).
+
+You can create your local registry to and your self-signed certificate at the
+same time:
```
mterhar, line 27:
```suggestion
To operate, Coder needs an image registry to store your Coder images.
Docker's `registry:2` image assumes that HTTPS will be used and supports
self-signed certificates.

Before starting the registry container, create a self-signed certificate at the
with a command like:
```

setup/air-gapped/infrastructure.md
```diff
@@ -0,0 +1,160 @@
+---
+title: Air-Gapped Network Setup
+description: Learn how to set up a network for air-gapped Coder deployment.
+---
+
+This article will walk you through setting up your network to support an
+air-gapped Coder deployment.
+
+If your network already has the following, you can proceed with your
+installation:
```
mterhar, line 10:
```suggestion
This article walks through setting up a the supporting infrastructure for
an air-gapped Coder deployment.

If the network already has the following, proceed with [the installation](../air-gapped):
```

mterhar:
@khorne3 may want to revet my last commit. I think it was misguided to undo the second-person.

khorne3 (approved):

PR #238, opened 2021-04-03
Add examples and behaviors to image tag docs

mterhar:
I included examples and things but the "summary" section
may be sufficient to get the point across.


khorne3:
@mterhar I felt the summary was a bit confusing without more context so I removed it. I don't feel too strongly about this, so if you'd like, I can re-add.

khorne3 (approved):

images/tags.md
```diff
@@ -38,3 +40,53 @@ To use an existing tag as the default tag, click the **vertical ellipses** for a
 tag and select **Make default**.
 
 ![Set Existing Tag as Default](../assets/existing-tag-as-default.png)
+
+## Image tag lifecycle
+
+Coder uses image tags to determine which image variant on which the user wants
```
jawnsy, line 46: This is a great improvement over our existing documentation, thanks for taking this on Mike! I think we should distinguish between three things:

1. What Coder does with tags (the refresh behavior, default tag handling)
2. How we recommend users use tags (note: I don't think the way we're currently using images/tags is ideal, for reasons I'll describe in an RFC)
3. How we currently use tags for our enterprise-images

A general challenge with Docker is that tags act similarly to Git's branches as well as tags (or, in other words, a git tag that is force-pushed). This means that the meaning of tags is arbitrary and varies according to what authors decide to do, though there are common practices that become de-facto standards (which, we're not currently following)

images/tags.md
```diff
@@ -38,3 +40,53 @@ To use an existing tag as the default tag, click the **vertical ellipses** for a
...
+  `ubuntu:20.04`), Coder will alert you regarding patches and security updates
+  so that you rebuild your environment (you won't get these fixes otherwise).
+  Coder does not, however, alert you regarding minor releases (e.g., movement
+  from `20.04` to `20.10`). This is a good option for those offering long-term
+  support of software with lengthier version cycles or those supporting multiple
+  versions where you expect to revert back to a prior release to investigate and
+  fix issues.
+
+- If you build your environment using a tag like
+  `codercom/enterprise-node:ubuntu`, the image name indicates the company and
```
jawnsy, line 82: The `codercom` organization contains all of our example images, `enterprise` refers to Coder Enterprise (and should probably be rebranded), and `node` refers to the fact that the image contains the build tooling necessary for building Node.js applications. The names are all arbitrary though.

images/tags.md
```diff
@@ -38,3 +40,53 @@ To use an existing tag as the default tag, click the **vertical ellipses** for a
...
+  software architecture (also known as the node), while the tag specifies the
+  base image. Without any additional information, such as a version number, you
+  can expect this to operate like a `rolling` or `latest` tag. You can also
```
jawnsy, line 85: It's never good to make assumptions - for example even with our enterprise images, we don't have Arch for anything except the base image. For example, our GoLand editor images only have CentOS and Ubuntu variants, not an Arch one: https://github.com/cdr/enterprise-images/tree/main/images/goland

I think we need to document this better with a table or something, though

images/tags.md
mterhar: I tried to provide some examples... people typically understand "latest" but we do some OS flavor and some version specific... Just re-reading this the use of 2nd person isn't good and the examples are contrived. 

It would be better if we just made a suggestion for how to use image tags and then let folks know they'd can do whatever they need.

images/tags.md
mterhar: This is one of those contrived examples. Definitely not trying to explain what we're doing but rather the outcome of what we've done.

mterhar: Furthermore, the example is actually bad because the enterprise-node:ubuntu image comes `FROM ubuntu:20.04` so it's not tied to rolling. It may roll but will require manual intervention.

images/tags.md
```diff
@@ -90,3 +99,36 @@ updates:
   associated with a specific project's major version. You can apply the `:v1`
   tag to the most recent build for the image, while you can use `:v1.3` or
   `:v1.3.1` to pull a more specific tag version.
+
+## Coder recommendation
```
mterhar, line 103: This whole section is a strawman that is intended to express the same thing as the "Latest versus numbered version" section... just that they should be deliberate and that it's flexible but also moving stuff around in unintuitive ways can slow the team or make excessive tag management.

PR #243, opened 2021-04-06
Org managers cannot create registries

mterhar:
Remove x from org managers -> create -> registries


mterhar:
It's probably worth doing a full review before merging this single character change.

mterhar:
Never mind... I switched orgs without noticing :-(

bpmct:
aw RIP. thanks for the effort @mterhar :)

Issue #244, opened 2021-04-06
Use node `12`


vapurrmaid:
- [ ] `@types/node`
- [ ] engines node version

PR #246, opened 2021-04-06
chore: Pin to node 12

vapurrmaid:
Context:

In #233 it was observed that we allow @types/node in LTS 14, but that
due to operational/architectural issues of cdr/docs and cdr/m, we should
instead consistently use version 12 of node.

Details:

- Set engines.node version to "^12.16.0". This is equivalent to the
  version used in cdr/m. Note that in npm, the caret means that the
latest minor and patch versions are OK - meaning 12.21.x is compatible
with ^12.16.0
- Use the latest version of @types/node, with a caret to indicate minor
  and patch versions are compatible

Resolves: #244

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "devDependencies": {
     "@types/minimist": "^1.2.1",
-    "@types/node": "^14.14.36",
+    "@types/node": "^12.20.7",
     "@types/semver-compare": "^1.0.1",
     "all-contributors-cli": "^6.20.0",
     "husky": "^6.0.0",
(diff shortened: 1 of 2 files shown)


jawnsy (approved):
:fire:

vapurrmaid:
PS I ran the changelog scripts to verify the new typings didn't create any runtime errors. All ran successfully :heavy_check_mark:

jawnsy:
@vapurrmaid might be nice to run the changelog scripts as part of an automated test, so that we can pick up problems in CI rather than requiring manual testing? just a thought :smile:

PR #263, opened 2021-04-09
chore: update GPUs info

khorne3:


khorne3:
@deansheather Underneath the text I updated, it says:

>By default, Coder allocates resources (CPU cores, memory, and disk space)
>based on the parent image.

Should I change this list of resources to include GPUs as well?

deansheather:
@khorne3 Images don't have GPUs as a default resource option, so we should leave it as is.

deansheather (approved):
Looks good but you should replace the picture with one that has a `0` in the GPU box. There was a PR that made sure there was a 0 there by default, but it hasn't been deployed yet.

vapurrmaid:
> Looks good but you should replace the picture with one that has a `0` in the GPU box. There was a PR that made sure there was a 0 there by default, but it hasn't been deployed yet.

We can easily do this in the web console - I'll message @khorne3 on Slack

vapurrmaid (approved):

PR #274, opened 2021-04-14
mention unsupported registries

bpmct:
AWS ECR: https://app.clubhouse.io/coder/story/2616/customer-amazon-elastic-container-registry-ecr-integration

GitHub Packages: https://app.clubhouse.io/coder/story/8094/use-tagged-images-and-avoid-imagepullpolicy-always

### DIFF
--- a/admin/registries/index.md
+++ b/admin/registries/index.md
@@ -40,7 +40,7 @@ select **Delete**.
 
 ## Unsupported registries
 
-The following registries do not behave as indended inside Coder at this time.
+Coder does not support the following registries at this time:
 
 - Amazon ECR
 - GitHub Packages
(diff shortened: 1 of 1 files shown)


admin/registries/index.md
```diff
@@ -37,3 +37,10 @@ When done, click **Import**.
 To delete a registry, go to the **Images** > **Registries** page. Find the
 registry that you'd like to remove, click its **horizontal ellipses** icon, and
 select **Delete**.
+
+## Unsupported registries
+
+The following registries do not behave as indended inside Coder at this time.
+
+- Amazon ECR
```
bpmct, line 45: I made a workaround for ECR can be periodically run every 10 hours to update the token in Coder with the API. Not sure if it is appropriate to mention here: https://github.com/bpmct/coder-ecr-sync

khorne3: I think if we're going to call it out as unsupported, we shouldn't link to the workaround. WDYT?

khorne3 (approved):

admin/registries/index.md
bpmct: Makes sense to me. The workaround is bad too.

PR #277, opened 2021-04-15
chore: add Action to check for old files

khorne3:


IAmHughes (approved):
Glad we got that working! This does include all the dirs we want to check, right?

.github/workflows/ci.yaml
```diff
@@ -39,3 +41,16 @@ jobs:
       - name: Formatting
         if: always()
         run: yarn format:check
+
+      - name: Find old files
+        uses: w3f/action-find-old-files@1.0.1
+        id: old_files
+        with:
+          minAge: 60
+          dirs: '["admin", "cli", "environments", "guides", "images", "setup"]'
```
vapurrmaid, line 50: I honestly believe that it may make sense to use `dirs: '["."]'` if that works, as I can see this getting out of date easily.

The other solution is to move all documentation into a root dir:

```
docs/
  admin/
  cli/
```

I won't request that change now as to keep us unblocked, but think we should switch soon

vapurrmaid (approved):

.github/workflows/ci.yaml
vapurrmaid: @IAmHughes - my proposal is we move all released MD files into a root path of its own (ex: `docs/`), as then this is a non-issue

khorne3: `dirs: '["."]'` worked perfectly so I did that

PR #282, opened 2021-04-19
MD Lint

khorne3:
- Brought over some of the existing rules we had for the old docs site
- Combined the config file with some of our old rules and our new ones (made some changes based on how we write stuff currently)
- Fixed linting errors

I also made a [chore to track upgrading markdown-lint](https://github.com/cdr/docs/issues/286).


package.json
```diff
@@ -29,6 +29,7 @@
     ],
     "*.md": [
       "markdownlint --fix",
+      "markdownlint --config .markdownlint-publicdocs.jsonc --rules .markdownlint-rules",
```
vapurrmaid, line 32: This file indicates some inconsistencies. In this branch, there are two `.markdownlint` configuration files:

- https://github.com/cdr/docs/blob/md-lint/.markdownlint.json
- https://github.com/cdr/docs/blob/md-lint/.markdownlint-publicdocs.jsonc

The only reason we'd do this is if we were linting different sections of the repository with different configurations. That was true back in `cdr/m`, but not true of `cdr/docs - as the invocation is using a `**/*.md` pattern, implying it runs on all markdown files in this repository.

Therefore we should:

1. combine the rules into one configuration file
2. remove `publicdocs` from the name of the configuration file

.markdownlint.json + .markdownlint-publicdocs.jsconc -> .markdownlint.jsonc

3. We should not invoke `markdownlint` twice in `lint-staged`, as that's quite inefficient. Simply add the `--fix` option here

vapurrmaid (approved):
LGTM!

PR #283, opened 2021-04-20
VNC: add local connection option and clarify web option

bpmct:
VNC can only be accessed via Dev URLs if the image includes both a VNC server (tigerVNC) and a web client (noVNC). 

If you only have a server on the image, or want a potentially faster connection, you can connect via SSH port forwarding.

### DIFF
--- a/guides/customization/vnc.md
+++ b/guides/customization/vnc.md
@@ -74,14 +74,15 @@ environment.
 
 1. Click **Create Environment**
 
-## Step 5: Create a Dev URL and access the VNC
+## Option 1: Connect via web
 
-Now that you've created your environment, you'll need to create a
-[dev URL](../../environments/devurls.md) so that you can access its services.
+If your image includes [noVNC](https://github.com/novnc/noVNC), or another
+web-based client, you can use a [dev URL](../../environments/devurls.md) to
+access it securely.
 
 1. From the **environment overview** page, click **Add URL**
 
-1. Provide the **Port** number that the VNC is running on (this information is
+1. Provide the **Port** number that noVNC is running on (this information is
    defined in the image you used to build this environment).
 
 1. Provide a **name** for the dev URL.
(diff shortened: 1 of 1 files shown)


bpmct:
also, forgot to mention: I tested VNC again with CVMs and did not run into an issue. I assume this was just due to some flakiness in our sample image, which is fine as it is just provided as a demonstration of how VNC can work in Coder

guides/customization/vnc.md
```diff
@@ -72,20 +72,17 @@ environment.
...
-Now that you've created your environment, you'll need to create a
-[dev URL](../../environments/devurls.md) so that you can access its services.
+If your image includes [noVNC](https://github.com/novnc/noVNC), or another
+web-based client, you can use a [dev URL](../../environments/devurls.md) to
+access it securely.
 
 1. From the **environment overview** page, click **Add URL**
 
-1. Provide the **Port** number that the VNC is running on (this information is
+1. Provide the **Port** number that noVNC is running on (this information is
```
ericpaulsen, line 85: I'm thinking we generalize this statement rather than specifically referencing noVNC, e.g:

> Provide the **Port** number that the VNC web-client is running on

ericpaulsen (changes requested):

guides/customization/vnc.md
khorne3: I like this so that the doc's more evergreen, though I understand we specifically use noVNC in the image

khorne3 (approved):

guides/customization/vnc.md
```diff
@@ -74,15 +74,22 @@ environment.
...
+### Option 1: Connect via web
 
 If your image includes [noVNC](https://github.com/novnc/noVNC), or another
 web-based client, you can use a [dev URL](../../environments/devurls.md) to
 access it securely.
 
 1. From the **environment overview** page, click **Add URL**
 
-1. Provide the **Port** number that noVNC is running on (this information is
+1. Provide the **Port** number that the VNC is running on (this information is
```
ericpaulsen: We should still specify that the user is accessing the VNC _web client_, to differentiate from the VNC _server_ used in the image as well.

ericpaulsen (changes requested):

ericpaulsen (approved):

PR #284, opened 2021-04-20
add port table

ericpaulsen:
adding this chart of all ports related to Coder resources. we've recently sent this to a couple of customers who wanted to hardwire their own network policies, so I thought it'd be helpful to be in docs for future cases.

initially, I intended on creating a markdown table replica of the chart, but had a hard time nesting multiple rows in the larger parent columns... so I just included the screenshot for PR to get this started. if someone could help me make one or add it themselves, I think it would look a bit cleaner.

including @f0ssel & @jawnsy to validate and add context if necessary.


khorne3:
@ericpaulsen Yeah, markdown tables aren't nice at all. I'll turn your screenshot into a table soon. Thanks!

setup/architecture.md
```diff
@@ -21,6 +21,17 @@ Coder deploys an NGINX Kubernetes ingress controller to allocate and route
 requests to the appropriate service. You can disable this controller in the helm
 chart if you use your ingress or gateway.
 
+## Ports
+
+Below is a table of the ports associated with Coder resources in the Kubernetes
```
jawnsy, line 26:
```suggestion
Below is a table of the listening ports associated with Coder resources in the
```

It might be worth indicating that these are inbound (listening) ports, and I don't think it's necessary to say "Kubernetes" here, we can just use "cluster"

Please also indicate in some way that these are listing our current ports and the list is subject to change

> IMO, it seems fair to document, provided that we indicate that these represent the current behavior and things may change over time. We would also need to consider how things change if we're using Istio or another service mesh

jawnsy (approved):
seems good to me provided that @khorne3 converts the image into a table :)

setup/architecture.md
ericpaulsen: Good catch - I will push a commit with that clarification. Thanks.

ericpaulsen:
> @ericpaulsen Yeah, markdown tables aren't nice at all. I'll turn your screenshot into a table soon. Thanks!

Awesome! I'll leave this PR open so you can push a commit with the table under this branch.

setup/architecture.md
```diff
@@ -21,6 +21,110 @@ Coder deploys an NGINX Kubernetes ingress controller to allocate and route
...
+      <tr>
+         <th>Pod</th>
+         <th>Source</th>
+         <th>Source port</th>
+         <th>Destination</th>
+         <th>Destination port</th>
+      </tr>
+      <tr>
+         <td rowspan="4">cemanager</td>
+         <td>loadbalancer</td>
```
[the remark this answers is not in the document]
khorne3: I checked cdr/m -- it's super inconsistent, but we should capitalize it the way it appears to the user

khorne3 (approved):

PR #288, opened 2021-04-21
add resource requirements and airgap notes

bpmct:


setup/kubernetes/local-preview.md
```diff
@@ -21,6 +21,11 @@ Before proceeding, please make sure that you have the following installed:
 
 ## Limitations
 
+### Resource requirements
+
+Make sure your host machine meets the [resource allocation minimums](../requirements)
+necessary to run Coder.
+
```
bpmct, line 28: @coadler if you have specifics here, that would be great!

coadler: fwiw, the script sets the resource requests for coder services to 0 and automatically sets cpu/mem overprovisioning high, plus the cluster setup is controlled by us, so I don't think much of this applies to the kind script. it might make more sense to mention the patform experience will directly be tied to system specs, and some performance drop is to be expected since it's all running inside a docker container.

setup/kubernetes/local-preview.md
```diff
@@ -54,6 +59,10 @@ environment.
 
 We are working on bringing Dev URL support to local previews in later releases.
 
+### Air-gapped Clusters
```
khorne3, line 62:
```suggestion
### Air-gapped clusters
```

setup/kubernetes/local-preview.md
```diff
@@ -54,6 +59,10 @@ environment.
...
+
+This option does not work in an air-gapped deployment.
```
khorne3, line 64:
```suggestion
The local preview option does not work in an air-gapped deployment.
```

setup/kubernetes/local-preview.md
khorne3: If you'd like to include Colin's caveat, maybe it could be worded as:

> Your experience with the local Coder preview is dependent on your system specs, but please note that you can expect slightly degraded performance due to the deployment running completely inside a Docker container.

And maybe the title would be "Resource allocation and performance" or something like that

khorne3 (approved):

khorne3:
Hey @bpmct is this good to go? If there are no objections, I'll merge today!

bpmct:
@khorne3 yep!

PR #171, opened 2021-03-17
Upgrade docs fixes (helm values, remove --force)

bpmct:
Some fixes based on my experience upgrading today.

PR #296, opened 2021-04-22
potential fix to workspace provider upgrade command

bpmct:
Context: I remember we removed `--force` from our standard update as it was constantly failing the update as described here: https://coder.com/docs/setup/updating#fixing-a-failed-upgrade: https://github.com/cdr/docs/pull/171

When updating my workspace provider, I ran into similar errors and my deployment succeeded without the command. I'm not exactly sure why, but I believe remember seeing `--force` only leads to a successful install on specific versions of helm. A sanity check would help, though.

### DIFF
--- a/admin/workspace-providers/deployment.md
+++ b/admin/workspace-providers/deployment.md
@@ -158,8 +158,7 @@ values changes; you can do this with
 helm upgrade coder-workspace-provider coder/workspace-provider \
     --version=[CODER_VERSION] \
     --atomic \
-    --install \
-    --force
+    --install
 ```
 
 If you want to update any of the helm chart's values, you can do so by supplying
--- a/workspaces/variables.md
+++ b/workspaces/variables.md
@@ -1,24 +1,23 @@
 ---
-title: Environment variables
-description:
-  Learn how to work with CODER_* environment variables inside workspaces
+title: Workspace variables
+description: Learn how to work with CODER_* workspace variables
 ---
 
-Coder injects a standard set of environment variables that allow you to access
+Coder injects a standard set of workspace variables that allow you to access
 contextual information about your workspace.
 
-To obtain a list of environment variables and their values, launch the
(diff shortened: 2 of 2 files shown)


vapurrmaid (commented):
Changes look good to me but @f0ssel should give the :heavy_check_mark:

f0ssel:
@sreya is this the general rec going forward? I can remove it in the coder-cli output as well

khorne3:
@sreya Do you mind taking a look at the questions in this PR, please? Thanks!

bpmct:
bump @sreya

workspaces/variables.md
```diff
@@ -1,23 +1,24 @@
 ---
-title: Workspace variables
-description: Learn how to work with CODER_* workspace variables
+title: Environment variables
```
jawnsy, line 2: I think these changes should be in a separate PR to keep the history clean

admin/workspace-providers/deployment.md
```diff
@@ -158,8 +158,7 @@ values changes; you can do this with
 helm upgrade coder-workspace-provider coder/workspace-provider \
     --version=[CODER_VERSION] \
     --atomic \
-    --install \
-    --force
+    --install
```
jawnsy, line 161: We no longer use `--force` for our internal deploys (preview, kind, dogfood, workspace providers) so I think it makes sense to update the docs accordingly, yeah.

sreya:
I guess this got lost in my barrage of emails but yes we no longer suggest using `--force`

workspaces/variables.md
bpmct: unsure how this got into here in the first place. thanks

deansheather (approved):

deansheather:
uh oh I didn't realize this was auto-merge 😬

bpmct:
all good @deansheather. i think we wanted to merge it.

PR #302, opened 2021-04-23
feat: connecting an external postgres

ericpaulsen:
adding docs for hooking up an external (to the cluster) postgres, as it is our recommended approach for production Coder deployments.

this is intended as an MVP, as I expect us to iterate on this as we continue to get customer feedback. open to thoughts & suggestions.

### DIFF
--- a/guides/deployments/postgres.md
+++ b/guides/deployments/postgres.md
@@ -1,84 +1,76 @@
 ---
-title: "Connect an External PostgreSQL"
+title: "PostgreSQL"
 description: "Learn how connect Coder to an external postgreSQL database."
 ---
 
-This guide walks through the steps to deploying Coder with an external
-PostgreSQL database connected.
+This guide walks you through deploying Coder with an external PostgreSQL
+database.
 
-## Connecting to an external PostgreSQL
+## Background
 
-By default, Coder will deploy a [TimescaleDB](https://www.timescale.com) inside
-the Kubernetes cluster used to install Coder. However, this is _only_
-recommended for evaluation purposes, as the database is not backed up, and can
-be lost if the cluster goes down.
+By default, Coder deploys a [TimescaleDB](https://www.timescale.com) inside the
+Kubernetes cluster to which you've installed Coder. However, we recommend this
+**only for evaluation purposes**, since the database isn't backed up and can be
+lost if the cluster goes down.
 
-For those reasons, we recommend bringing in your own PostgreSQL for production
-deployments, and hosting it external to the cluster. Below are the steps to do
-so:
+As such, we strongly recommend using a PostgreSQL database for production
+deployments and hosting it **outside** the Kubernetes cluster hosting Coder.
 
-1. Spin up a PostgreSQL instance if you have not already done so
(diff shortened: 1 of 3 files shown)


guides/deployments/postgres.md
```diff
@@ -0,0 +1,109 @@
...
+- This private IP will be referenced in the Coder helm chart configuration
+
+1. If your PostgreSQL instance has a password, follow the below:
+
+- Open the terminal, connect to your cluster and create a secret:
+
+  ```console
+  kubectl create secret generic <YOUR-DB-SECRET> \
+      --from-literal=password=<YOUR-DATABASE-PASSWORD> \
+  ```
```
sreya, line 33: This is fine but we should encourage people to create passwords in a way where the secret isn't recorded in your shell history. For example this is a more security-conscious way to create a secret on the cmd line 
```bash 
kubectl create secret generic <NAME> --from-file=test=/dev/stdin
```

guides/deployments/postgres.md
```diff
@@ -0,0 +1,109 @@
...
+By default, Coder will deploy a [TimescaleDB](https://www.timescale.com) inside
+the Kubernetes cluster used to install Coder. However, this is _only_
+recommended for evaluation purposes, as the database is not backed up, and can
+be lost if the cluster goes down.
+
+For those reasons, we recommend bringing in your own PostgreSQL for production
+deployments, and hosting it external to the cluster. Below are the steps to do
+so:
+
+1. Spin up a PostgreSQL instance if you have not already done so
```
sreya, line 20: This step should be the bulk of the guide. This page does not otherwise provide much beyond the existing documentation for referencing a postgres instance in the helm.

sreya (commented):
I'll leave others to review grammar but content-wise this seems ok. I think for this guide to be valuable it should mainly be about how to setup a Postgres instance. We already have documentation for how to reference postgres in the helm [here](https://coder.com/docs/setup/installation#installing-coder) on step 4.

ericpaulsen:
> I'll leave others to review grammar but content-wise this seems ok. I think for this guide to be valuable it should mainly be about how to setup a Postgres instance. We already have documentation for how to reference postgres in the helm [here](https://coder.com/docs/setup/installation#installing-coder) on step 4.

Good point - since there are many ways someone can setup a postgres, should we focus on documenting how to set up a postgres within each cloud provider?

guides/deployments/postgres.md
ericpaulsen: I can update the sample command in [Step 4](https://coder.com/docs/setup/installation#installing-coder).

khorne3:
@ericpaulsen Do you think it makes sense to link to the tutorials instead (e.g., [GKE's PostgreSQL docs](https://cloud.google.com/architecture/deploying-highly-available-postgresql-with-gke))?

ericpaulsen:
> @ericpaulsen Do you think it makes sense to link to the tutorials instead (e.g., [GKE's PostgreSQL docs](https://cloud.google.com/architecture/deploying-highly-available-postgresql-with-gke))?

Yes, as long as we include links for each of the 3 major cloud providers.

khorne3 (approved):

PR #312, opened 2021-04-30
chore: modify K8s requirements

khorne3:
Update suggested resource allocations for Coder deployments

### DIFF
--- a/setup/requirements.md
+++ b/setup/requirements.md
@@ -14,12 +14,15 @@ You will also need to allocate additional CPU cores, RAM, and storage for
 will vary based on your project, though you can use the following sets of
 minimum requirements to help you estimate:
 
-- Development using a JetBrains IDE: 8 GB RAM, 4 cores, 10 GB Disk per developer
-- Development using VS Code with an SSH connection to Coder: 1 GB RAM, 1 core
-  per developer
+- Development using a JetBrains IDE: 4 CPU cores, 8 GB of RAM, and 10 GB of
+  storage per developer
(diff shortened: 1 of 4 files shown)


setup/requirements.md
```diff
@@ -6,8 +6,10 @@ description: Learn about the prerequisite infrastructure requirements.
...
 
-For **basic control services**, allocate at least 2 CPU cores, 4 GB of RAM, and
-20 GB of storage.
+For **basic control services**, allocate at least 8 CPU cores, 32 GB of RAM, and
+50 GB of storage (if you later determine via [usage
+monitoring](../guides/admin/usage-monitoring.md) that these are too high/low,
+you can change the allocation).
 
 For **each** active developer, please allocate an additional CPU core, 1 GB of
 RAM, and 10 GB of storage on top of the resources allocated for basic control
```
bpmct, line 15: I think _basic control services_ refers to the things that run in a Coder cluster, regardless of user count and workspaces.

This section, however, makes a lot of assumptions. For example, if an "active developer" needs to use JetBrains, they need 8 GB RAM and 4 CPU cores. If a user is doing more advanced development on VS Code, this recommendation is still not high enough. This could perhaps be rephrased to:

> Additionally, you will need to allocate space on your cluster for **each** active developer. This will vary depending on your team's resource workloads and should be adjusted as needed. See the minimum requirements below for reference:
- Using JetBrains Development: of 8 GB RAM, 4 cores, 10 GB Disk per developer
- VS Code and SSH: 1 GB RAM, 1 core per developer

bpmct (approved):
This is huge, thanks!!

PR #315, opened 2021-05-03
chore: suppress node version changes by dependabot

khorne3:


jawnsy (approved):
Thanks for taking care of this! And thanks @jsjoeio for giving us the delicious copypasta :spaghetti:

PR #319, opened 2021-05-04
Explain install.sh must be executable

brianbegy:
Add a sentence explaining that the install.sh must be marked executable and that coder won't do it for you.

I was assuming that coder would mark as executable anything it intended to execute.  That is not the case.

### DIFF
--- a/workspaces/personalization.md
+++ b/workspaces/personalization.md
@@ -98,7 +98,9 @@ workspace or turn it on.
 
 At startup, Coder clones your dotfiles repository into `~/dotfiles`. If there's
 an executable `~/dotfiles/install.sh` present, Coder executes it. If not, all
-dot-prefixed files are symlinked to your home directory.
+dot-prefixed files are symlinked to your home directory.  
+
+You must mark the install.sh file as executable before committing it to your dotfiles repo.
 
 Read more about dotfiles repos [here](http://dotfiles.github.io/).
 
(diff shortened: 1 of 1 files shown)


workspaces/personalization.md
```diff
@@ -98,7 +98,9 @@ workspace or turn it on.
 
 At startup, Coder clones your dotfiles repository into `~/dotfiles`. If there's
 an executable `~/dotfiles/install.sh` present, Coder executes it. If not, all
-dot-prefixed files are symlinked to your home directory.
+dot-prefixed files are symlinked to your home directory.  
+
+You must mark the install.sh file as executable before committing it to your dotfiles repo.
```
IAmHughes, line 103:
```suggestion
- _Note: You must mark the `install.sh` file as executable before committing it to your dotfiles repo._
```

Just a minor suggestion on formatting

IAmHughes (changes requested):
@khorne3 let me know if you prefer this format. I think we should at least put `install.sh` in code tags but I'm not convinced on the `Note` prefix I added

workspaces/personalization.md
brianbegy: That makes sense.

khorne3 (approved):

vapurrmaid:
@all-contributors please add @brianbegy  for content

PR #320, opened 2021-05-04
fix occurrences of "workspace variables" -> "environment variables

bpmct:


IAmHughes (approved):
Good catch, probably happened in the rename of environments => workspaces but should not have been changed for env vars.

khorne3 (approved):

PR #321, opened 2021-05-04
set AWS cluster variables in separate command

bpmct:
With some shells, variables cannot be directly set in the same command

see: https://gist.github.com/bpmct/f4b697c1bd2861f8a5d6766897ace8e1


IAmHughes (approved):
Good call out :shipit:

khorne3 (approved):

PR #322, opened 2021-05-04
add one-liner for password reset

bpmct:
This is also a more-reliable way to get the initial admin password, as it can sometimes be lost in the logs after an install. A one-liner makes it easier to quickly fetch the password, while still documenting what is being ran.


khorne3:
@bpmct What are your thoughts on only providing the one-line option? Are there upsides to providing both options?

bpmct:
> @bpmct What are your thoughts on only providing the one-line option? Are there upsides to providing both options?

I like how each step is explained, so that if the command fails, the user knows what each step is attempting to do.

We can probably just keep the one-liner as long as we explain what the external and internal command does.

khorne3 (approved):

PR #326, opened 2021-05-05
Correct `gpu-count` on WAC templates

IAmHughes:
The current `gpuCount` is not correct and it should be `gpu-count`.

Asking @Emyrk to review as we may be making additional changes to stay consistent, but for now I want to verify that I have things right in the template 👍

### DIFF
--- a/workspaces/workspaces-as-code/templates.md
+++ b/workspaces/workspaces-as-code/templates.md
@@ -37,7 +37,7 @@ workspace:
     cpu: 4
     memory: 16
     disk: 128
-    gpuCount: 1
+    gpu-count: 1
     labels:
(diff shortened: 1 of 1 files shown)


Emyrk (approved):
All consistency changes will happen when we migrate from 0.1 -> 0.2. 

That change changes the entire layout of the yaml, so it's a good time to do a full consistency check.

Emyrk:
@khorne3 This is good to merge from a technical perspective. It affects the current release (1.18). Unsure if I'm allowed to click :smile:

khorne3:
@Emyrk you are, but I'll take care of it this time 'round -- thank you! :)

PR #328, opened 2021-05-05
add: logging docs

ericpaulsen:
submitting this to flesh out our current logging documentation included in `values.yaml` and add it to `docs` for greater visibility & ease of use for customers.

I assume `/dev/stderr` is the typical file path used for storing logs, but I'd like some validation as to whether this is truly the case, or if users can define any file path they'd like.


setup/logging.md
```diff
@@ -0,0 +1,83 @@
...
+
+This is the default value set in the Helm chart, as reflected below:
+
+```yaml
+logging:
+    human: /dev/stderr
+```
+
+When set, logs will be sent to the `/dev/stderr` file path, and formatted for human
+readability.
```
bpmct, line 24: This may be a noob question, but how would someone fetch these with `kubectl`? Might help to include this here.

bpmct (approved):

setup/logging.md
ericpaulsen: Agreed. This I'm not sure of either. Would like some clarification here.

coadler: `/dev/stdout` & `/dev/stderr` both write to the pods stdout/stderr, accessible via `kubectl logs <podname>`

khorne3:
I thought this fit better under guides, but it might be helpful for people to review their logging settings when getting started, so I added a note on the Install guide pointing to this page.

khorne3 (approved):

PR #335, opened 2021-05-11
chore: Organizations docs

khorne3:
We didn't really have anything on orgs, so I created a new doc for this. The current doc (which lives under access control) will focus only on org roles.

### DIFF
--- a/admin/organizations.md
+++ b/admin/organizations.md
@@ -14,21 +14,21 @@ are.
 
 ## The default organization
 
-When you first set up Coder, you'll generate the default organization. You can
-then assign users and their workspaces to that organization.
+Coder automatically creates a default organization for you during the deployment
+process. You can then assign users and their workspaces to that organization.
 
-There must always be a default organization, but you can change the one set as
-the default once you have two or more organizations.
+If you have multiple organizations, you can set one or more as the default; you
+can also change which organizations are defaults at any time.
 
 ## Create a new organization
 
-Site admins and site managers can create new organizations by going to
-**Manage** > **Organizations** > **New Organization**.
+[Site admins and site managers](access-control/user-roles.md) can create new
+organizations by going to **Manage** > **Organizations** > **New Organization**.
 
 ![Create a new organization dialog](../assets/create-an-org.png)
 
 Provide a **name** and (optionally) a **description** for this organization. If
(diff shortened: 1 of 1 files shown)


vapurrmaid:
@khorne3 I honestly don't mind targeting `master` for this so your work isn't squashed down into the changelog release. It seems like this can be released at any time, not 1.20 only.

Are you OK if I rebase this onto `master` (up to you). It seems like there's some other commits in there

khorne3:
@vapurrmaid Changed base branch back to master -- this should be ok

admin/organizations.md
```diff
@@ -0,0 +1,86 @@
...
+end-user can only access images that are assigned to the same organization they
+are.
+
+> Be sure to familiarize yourself with the [types of
+> roles](access-control/organizations.md) you can assign users within an
+> organization.
+
+## The default organization
+
+When you first set up Coder, you'll generate the default organization. You can
```
vapurrmaid, line 17: Maybe

> A default organization will automatically be created for you.

Or something along those lines, since it doesn't require a user to invoke an action to cause the generation

admin/organizations.md
```diff
@@ -0,0 +1,86 @@
...
+then assign users and their workspaces to that organization.
+
+There must always be a default organization, but you can change the one set as
+the default once you have two or more organizations.
```
vapurrmaid, line 21: There can be more than one default. So there must always be at least one, and you can set/unset which one(s) are default.

admin/organizations.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Create a new organization
+
+Site admins and site managers can create new organizations by going to
```
vapurrmaid, line 25: should we link to user roles here, or nah?

admin/organizations.md
```diff
@@ -0,0 +1,86 @@
...
+**Manage** > **Organizations** > **New Organization**.
+
+![Create a new organization dialog](../assets/create-an-org.png)
+
+Provide a **name** and (optionally) a **description** for this organization. If
+you want this to become your **Default organization**, make sure to check the
```
vapurrmaid, line 31:
```suggestion
you want this to become a **Default organization**, make sure to check the
```

vapurrmaid (approved):

PR #340, opened 2021-05-12
chore: add info re: admin permission level setting for dev URLs

khorne3:


admin/devurls.md
```diff
@@ -51,3 +51,45 @@ The final step to enabling dev URLs is to update your wildcard DNS record. Get
...
+    <td>Accessible by anyone (no authentication required)</td>
+  </tr>
+  <tr>
+    <th>Authenticated</th>
+    <td>Accessible by any authenticated Coder user</td>
+  </tr>
+  <tr>
+    <th>Organization</th>
+    <td>Accessible by anyone in the user's organization</td>
+  </tr>
```
vapurrmaid, line 77: I don't know how to gracefully mention this, but it's true that site managers and the site admin can still access them AFAICT, because those roles have elevated permissions that let them transiently access any organization.

I don't know if that information is bet called out in the user roles doc, or other spots like this one.

What do you think?

vapurrmaid (commented):
Looks great - just a question/observation that I'm unsure about!

admin/devurls.md
khorne3: I think this is the only place in the UI where the end-user can change the permissions level of something, right? Everything else requires elevated permissions, I think. So, I think a small banner that says

> Site admins and site managers can access any dev URL in the Coder deployment

would make this fact clear to the user.

vapurrmaid: That's an awesome point

vapurrmaid: So we'll hold off on this suggestion for now due to https://app.clubhouse.io/coder/story/12644/site-managers-cannot-access-organization-and-private-dev-urls

vapurrmaid (approved):

PR #346, opened 2021-05-13
add storage & db requirements

ericpaulsen:
adding storage requirement info, specifically around the use of PVCs and block storage. also reiterating our recommendation of using an external PostgreSQL instance for production deploys, and noting the minimum version required.

open to feedback & word-smithing.

### DIFF
--- a/setup/requirements.md
+++ b/setup/requirements.md
@@ -64,12 +64,17 @@ the block storage requested.
 ## Database
 
 Coder requires the use of a [PostgreSQL](https://www.postgresql.org) database to
-store metadata related to your deployment. By default, Coder will deploy a TimescaleDB
+store metadata related to your deployment.
+
+By default, Coder will deploy a TimescaleDB
 internal to your Kubernetes cluster. This is included for evaluation purposes _only_,
 as it is not backed up. For production deployments, we recommend using a PostgreSQL
 database _external_ to your cluster. You can connect Coder to your external database
 by [modifying the Helm chart](https://coder.com/docs/guides/admin/helm-charts) with
 the appropriate PostgreSQL values.
+
+If providing your own instance, a minimum version of PostgreSQL 11 is required with
+the `contrib` package installed.
 
 ## Network Policies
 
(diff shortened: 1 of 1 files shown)


khorne3 (approved):

setup/requirements.md
```diff
@@ -53,6 +53,29 @@ currently require the following versions _or newer_:
...
 launches the Remote IDE in a pop-up window.
 
+## Storage
+
+Coder requires the use of a [persistent
+volume](https://kubernetes.io/docs/concepts/storage/persistent-volumes/) in your
+Kubernetes cluster to store [workspaces](../workspaces/index.md) data. More
+specifically, the persistent volume claim (PVC) requires the block storage type
+(the PVC is created when you create the workspace to mount the requested block
+storage).
```
bpmct, line 63: @ericpaulsen sorry to reopen an old discussion I must have missed, but I remember seeing this in our community slack regarding a bare-metal install:

> It looks like you're missing a dynamic volume provisioner, which I can imagine happening easily if you're running "on bare metal". See https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/ for what that's about, but see https://kubernetes.io/docs/concepts/storage/storage-classes/ for the main documentation on how to fix it. You'll likely want to set a "default" storage class (see https://kubernetes.io/docs/tasks/administer-cluster/change-default-storage-class/), unless you have particular needs, whereupon you'll need to tweak the Coder helm chart to set the storage class on the generated PVCs. (edited) 

Asking because I'm genuinely not sure - does this description infer the use of a Dynamic Volume Provisioner?

ericpaulsen: @bpmct I _think_ Coder assumes the use of Dynamic Volume Provisioning (per the definition), but I'm not entirely sure. I'm waiting on some feedback from product to validate. I'll be sure to update docs accordingly.

ericpaulsen: Can confirm that Coder assumes the use of Dynamic Volume Provisioning to properly mount PVCs & create workspaces. I'll note this in a new PR.

PR #352, opened 2021-05-18
Configuring external database validation

mterhar:
In the process of making the terraform stuff, some DB config examples need adjustment.


jawnsy (approved):
LGTM. It's kind of surprising that `kubectl` doesn't have a `--from-stdin` option or accept `--from-file=test=-`, because this instruction will only work on Linux

khorne3:
@mterhar Is this still WIP or good to go?

PR #354, opened 2021-05-19
add steps for installing coder in an archive

ericpaulsen:
adding steps for pulling down the helm chart & installing Coder in an archive, as a customer has run into this issue. @sreya provided the steps (included in this PR) as a solution to the customer.

### DIFF
--- a/guides/deployments/archive-install.md
+++ b/guides/deployments/archive-install.md
@@ -1,0 +1,29 @@
+---
+title: Install Coder from an archive
+description: Learn how to install Coder from an archive
+---
+
+You can install Coder from an archive instead of using Helm. To do so, replace
+**steps 1-2** of the [Installation guide](../../setup/installation.md) with the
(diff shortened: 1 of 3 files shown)


khorne3:
@ericpaulsen How common is this? I'm wondering if we should turn this into a brief guide and link out to it?

ericpaulsen:
> @ericpaulsen How common is this? I'm wondering if we should turn this into a brief guide and link out to it?

Not very common from my personal experience. Since it's only a few steps I think we can just keep it in the install steps for easy reference.

khorne3:
I think, given that the install process is challenging enough as is, we should move it out -- it's short, but I could see this being confusing for people. I'd be happy to do this

ericpaulsen:
That sounds good with me. Thanks for taking it on.

khorne3 (approved):

PR #355, opened 2021-05-19
Add netv2 changes

f0ssel:
Deciding to omit a dedicated doc page for this feature since we pretty much just expect pilot customers to enable these options as validation for us more than anything and we have a whole dedicated doc for them. This is open to discussion if someone feels strongly otherwise.

### DIFF
--- a/changelog/1.19.0.md
+++ b/changelog/1.19.0.md
@@ -9,7 +9,7 @@ description: "Released on 05/19/2021"
   `/opt/coder`, such as code-server and coder-cli, have been moved to
   `/var/tmp/coder`
 - infra: Workspaces as code templates have been updated from version _0.1_ to
-  _0.2_. Version _0.1_ templates will no longer work. 
+  _0.2_. Version _0.1_ templates will no longer work.
 - web: The workspace as code _Open in Coder_ embeddable button flow no longer
   includes a _clone_ step; buttons created from prior versions will still work,
   but Coder will not clone the project. To automate project cloning, use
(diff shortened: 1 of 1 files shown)


changelog/1.19.0.md
```diff
@@ -18,6 +18,13 @@ description: "Released on 05/19/2021"
   **Admin** > **Templates** no longer uses the following fields: **Project
   Repository URL** and **Project Git Service**
 
+### Deprecation notices 
+
+- helm: The `dashboard` deployment and service are being removed in a future
+  release. Any customers relying on routing rules outside the default ingress
+  included in the helm chart should remove any routing rules to this service
+  and direct traffic to the `cemanager` service instead.
```
kylecarbs, line 26: Did the embed stuff get in to 1.19? Otherwise IDK if we can deprecate this

f0ssel: Lol I should be asking you that question

kylecarbs: There were a few embed things fixed post-this... so I don't think we should deprecate here.

vapurrmaid: IMO we should close this PR (despite the beautiful CLOG additions) due to this:

https://www.notion.so/How-To-Ship-Customer-Visible-Changes-477b63d362aa400ba9075a01b09084bd

vapurrmaid: LMK if I'm off-base with that thought - but basically this isn't something we document until beta stage

vapurrmaid (commented):
I'm currently of the opinion we won't be listing bleeding edge features that aren't enabled and have no behavioral effect without manual opt-in at a low-level until they're in `beta`

f0ssel:
fine with that decision, we can always add something if someone else cares

PR #374, opened 2021-05-26
feat: azure DNS config & ssl setup via cert-manager

ericpaulsen:
docs for setting up SSL certificates using Azure DNS & cert-manager. Referenced documentation includes:

- [cert-manager Azure DNS docs](https://cert-manager.io/docs/configuration/acme/dns01/azuredns/)
- [aad-pod-identity docs](https://azure.github.io/aad-pod-identity/docs/demo/standard_walkthrough/)

please wordsmith & format where necessary.

### DIFF
--- a/guides/ssl-certificates/azureDNS.md
+++ b/guides/ssl-certificates/azureDNS.md
@@ -21,8 +21,8 @@ cert-manager:
 - [Managed Identity Using AKS Kubelet Identity](https://cert-manager.io/docs/configuration/acme/dns01/azuredns/#managed-identity-using-aks-kubelet-identity)
 - [Service Principal](https://cert-manager.io/docs/configuration/acme/dns01/azuredns/#service-principal)
 
-This guide will only walk through the _first_ option. Yet the prerequisites will
-remain the same across all three.
+This guide will only walk through the **first** option, though the prerequisites
+are the same regardless of which option you choose.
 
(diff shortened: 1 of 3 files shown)


bpmct:
I don't use Azure often, so I'd probably be a good subject to test this. Super pumped to see more DNS providers being added.

ericpaulsen:
Awesome! That'd be great if you can dogfood and leave feedback.

jawnsy:
@lilshoff is doing some work to automate certificate provisioning using LetsEncrypt, in a manner that is built-in to Coder, which seems like it'd make this redundant.

@ketang @nathanpotter I don't know what our backward compatibility guarantees should be with respect to our documentation, or if we need to think about a migration story once we merge Lily's TLS work.

I don't think it's a big deal for customers to stop using cert-manager and start using our built-in support instead, so maybe breaking compatibility is OK or maybe we don't consider this breaking compatibility? In that case, we could merge this and remove it later, once Lily's TLS work is available.

Alternatively, we could continue supporting this configuration indefinitely (cert-manager in front of Coder). Thoughts/comments?

ericpaulsen:
@jawnsy - docs versioning is WIP at the moment, so I think we can merge this and keep it as a part of docs version 1.19. Once we release automated certificate provisioning, we can revisit this and remove if deemed redundant.

PR #378, opened 2021-05-28
chore: replace architecture diagram

khorne3:
[Preview](https://codercom-f5v54ntwg-codercom.vercel.app/docs/setup/architecture)


jawnsy:
Ah, the image is kinda hard to read in the docs due to scaling, I think we're gonna need some help making one for these dimensions.

Also, we list dashboard, envproxy, etc twice - presumably it's because we're trying to indicate the pod (outer) as well as container (inner) but I don't think that distinction matters for an architecture diagram like this.

Wondering if "Internet" makes sense here, or if we want to use a term like "Internet/Intranet" or just de-emphasize the outside (the point is that it's traffic _outside_ Coder being routed _into_ Coder) -- for offline installations, there's no inbound/outbound Internet access

Instead of "TCP load balancer", we might want to use "Kubernetes LoadBalancer Service" to clarify what it is, rather than being a generic load balancer or implying that it's something a user needs to supply

bpmct:
@jmilling33 would likely be the one making these changes. He's working on a new diagram brand for Coder

PR #400, opened 2021-06-10
Add note about accessurl being https

Emyrk:
Associated with https://github.com/cdr/m/pull/8659

### DIFF
--- a/admin/access-url.md
+++ b/admin/access-url.md
@@ -48,3 +48,5 @@ helm upgrade coder coder/coder \
 
 1. Provide your custom domain in the **Access URL** field. The URL you provide
    must match the value you provided as `ingress.host` in the previous step.
+   
+> If your Access URL uses the HTTPS protocol, you will not be able to access your deployment over HTTP via the browser. 
--- a/admin/access-url.md
+++ b/admin/access-url.md
@@ -49,4 +49,4 @@ helm upgrade coder coder/coder \
 1. Provide your custom domain in the **Access URL** field. The URL you provide
    must match the value you provided as `ingress.host` in the previous step.
    
-> If your Access URL uses the HTTPS protocol, you will not be able to access your deployment over HTTP via the browser. 
+> If your Access URL uses the HTTPS protocol, non-secure connections (HTTP) will be unable to authenticate with your deployment.


admin/access-url.md
```diff
@@ -48,3 +48,5 @@ helm upgrade coder coder/coder \
 
 1. Provide your custom domain in the **Access URL** field. The URL you provide
    must match the value you provided as `ingress.host` in the previous step.
+   
+> If your Access URL uses the HTTPS protocol, non-secure connections (HTTP) will be unable to authenticate with your deployment.
```
jawnsy, line 52: What should a user do instead? It seems in this case, can we automatically redirect from http to https?

Showing an insecure login page that won't work seems like bad UX to me, documentation is better than nothing but fixing the problem in code seems ideal.

Do you know if nginx gives us the Forwarded header that would allow us to detect that the end user is on HTTP? If the scheme is http and the Forwarded doesn't say that the user actually came in through HTTPS (i.e. a reverse proxy is terminating TLS for us), then could we log a warning or something? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded

Emyrk: Yea, I'm trying to get some consensus on how to handle this. I really don't know the best approach. 

I'd like to automatically redirect users, but they might never hit "proceed with caution" in the error case to fix the settings. There is a way to detect this from the FE if the BE sends over a custom header that indicates it's using a TLS deployment. Since the BE has the accessURL.

I don't want to rely on nginx. And I don't know enough about different proxy deployments to know if I can rely on any singular header.

---

This `secure` header **will** break someone who is attempting to downgrade to http. The http site cannot access the `secure` cookie, so it cannot delete it or read it. It is stuck until it goes to `https` to logout, or manually deletes the cookies in their browser settings.

jawnsy: I guess we could rely on JavaScript then, and show a warning in the frontend perhaps? Is there some way to detect that the server is returning Secure cookies while the frontend is accessed over HTTP?

Emyrk: My best thought on how to do that is to send over a custom header. Like `X-Coder-TLS: true` or something.
There is no way for the FE to read the cookie, overwrite the cookie, or do anything with the cookie. So they would have to infer it from a failed login attempt. Which also kinda sucks.


This should be a rare occurance? Like who downgrades from https -> http? If they do, we'll likely get a support ticket, and hence I added that note in the docs.

jawnsy: Yeah, that seems fair enough that it'll be a rare occurrence and maybe not worth solving now. I think it'd be easier for us to do something different once we remove the ingress anyways, it simplifies things for us a bit

jawnsy (approved):

admin/access-url.md
Emyrk: If we do all TLS in the app, we would know 100% if we can do redirects and everything more safely

Emyrk:
Talked to Ketan, went with a toggle for now.

PR #401, opened 2021-06-10
Avoid pulling full values file during install

deansheather:
Pulling full values file from the helm chart has stuff like `cemanager.image` etc. which we don't want the user to have in their values file unless they are specifically pulling images from another registry.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "format:check": "prettier --check '**/*.{css,html,js,json,jsx,ts,tsx,yaml,yml}'",
     "format:write": "prettier --write '**/*.{css,html,js,json,jsx,ts,tsx,yaml,yml}'",
     "lint": "markdownlint --config .markdownlint.jsonc --rules .markdownlint-rules '**/*.md'",
-    "lint:fix": "markdownlint --fix '**/*.md'"
+    "lint:fix": "markdownlint --config .markdownlint.jsonc --rules .markdownlint-rules --fix '**/*.md'"
   },
   "devDependencies": {
     "@types/minimist": "^1.2.1",
--- a/setup/installation.md
+++ b/setup/installation.md
@@ -46,13 +46,13 @@ kubectl config set-context --current --namespace=coder
    `helm search repo coder -l`)
 
    ```console
-   helm install coder coder/coder -n coder --version=<VERSION>
+   helm install coder coder/coder -namespace coder --version=<VERSION>
    ```
 
    **Steps 3-5 are optional for non-production deployments.**
 
 1. Get a copy of your Helm config values so that you can modify it; you'll need
-   to modify the Helm chart to update your PostgreSQL databases (step 4) and
+   to modify these values to update your PostgreSQL databases (step 4) and
    enable dev URLs (step 5):
 
    a. Get a copy of your existing Helm values and save it as `values.yaml`:
@@ -60,8 +60,8 @@ kubectl config set-context --current --namespace=coder
 
    b. Edit the `values.yaml` file as needed.
 
-   > View the
-   > [configuration options available in the `values.yaml` file.](https://github.com/cdr/enterprise-helm#values)
+   > View the [configuration options available in the `values.yaml`
+   > file.](https://github.com/cdr/enterprise-helm#values)
 
    c. Upgrade/install your Coder deployment with the updated Helm chart (be sure
    to replace the placeholder value with your Coder version). **This must be
--- a/setup/installation.md
+++ b/setup/installation.md
@@ -46,7 +46,7 @@ kubectl config set-context --current --namespace=coder
    `helm search repo coder -l`)
 
    ```console
-   helm install coder coder/coder --namespace coder
+   helm install coder coder/coder -n coder --version=<VERSION>
    ```
 
    **Steps 3-5 are optional for non-production deployments.**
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
     "format:check": "prettier --check '**/*.{css,html,js,json,jsx,ts,tsx,yaml,yml}'",
     "format:write": "prettier --write '**/*.{css,html,js,json,jsx,ts,tsx,yaml,yml}'",
     "lint": "markdownlint --config .markdownlint.jsonc --rules .markdownlint-rules '**/*.md'",
-    "lint:fix": "markdownlint --fix '**/*.md'"
+    "lint:fix": "markdownlint --config .markdownlint.jsonc --rules .markdownlint-rules --fix '**/*.md'"
```
jawnsy, line 15: nice fix :+1:

setup/installation.md
```diff
@@ -46,39 +46,37 @@ kubectl config set-context --current --namespace=coder
    `helm search repo coder -l`)
 
    ```console
-   helm install coder coder/coder --namespace coder
+   helm install coder coder/coder -n coder --version=<VERSION>
```
jawnsy, line 49: the long forms were intentional, because they're more readable than `-n` - can we change this back?

setup/installation.md
```diff
@@ -46,39 +46,37 @@ kubectl config set-context --current --namespace=coder
...
-   will override those in the default chart.
+   b. Edit the `values.yaml` file as needed.
 
    > View the
    > [configuration options available in the `values.yaml` file.](https://github.com/cdr/enterprise-helm#values)
 
    c. Upgrade/install your Coder deployment with the updated Helm chart (be sure
-   to replace the placeholder value with your Coder version). **This must be done
-   whenever you update the Helm chart:**
+   to replace the placeholder value with your Coder version). **This must be
```
jawnsy, line 67: suggest adding a few spaces to align "to" with "Upgrade"

setup/installation.md
```diff
@@ -96,14 +94,14 @@ kubectl config set-context --current --namespace=coder
    ```
 
    To create the `passwordSecret`, run
-   `kubectl create secret generic <NAME> --from-file=test=/dev/stdin`
+   `echo "UserDefinedPassword" | kubectl create secret generic <NAME> --from-file=password=/dev/stdin`
```
jawnsy, line 97: This defeats the purpose... the reason this was suggested before was to avoid putting secrets in history files (the same can be achieved by prefixing the command with a space, of course). Doing this echo puts the password back in the history

jawnsy (approved):

setup/installation.md
deansheather: I changed it to match the upgrade command below this, i'll change both though

setup/installation.md
deansheather: If we're saying to use stdin in a terminal then they have to type the password and hit `Ctrl+D`, which we don't tell people to do right now

deansheather: So really it's a balance between making the installation experience easy and risk logging password, or making the installation experience more complicated: "Oh just run this command, type your password then if you're on windows click ctrl+d or if you're on mac click cmd+d..."

I think we should use from-literal and put a warning saying it may be logged in shell history, and if they don't want that then put a space at the front

jawnsy: I think @sreya suggested this here: https://github.com/cdr/docs/pull/302#discussion_r619551224, you might want to check with him. I do not have strong feelings about it. It's too bad kubectl doesn't have something built-in to do this.

This way of doing it probably doesn't work on Windows, right? I also dunno about mac, presumably it's Unix-y enough that it works there? :man_shrugging:

setup/installation.md
```diff
@@ -46,39 +46,37 @@ kubectl config set-context --current --namespace=coder
...
 
-   a. Get a copy of your existing Helm chart and save it as `current-values.yaml`:
-   `helm show values coder/coder > current-values.yaml`
+   a. Get a copy of your existing Helm values and save it as `values.yaml`:
+   `helm get values coder -n coder | tail -n +2 > values.yaml`
 
-   b. Edit the `current-values.yaml` file as needed. Be sure to remove the lines
-   that you are _not_ modifying, otherwise the contents of `current-values.yaml`
-   will override those in the default chart.
+   b. Edit the `values.yaml` file as needed.
```
ericpaulsen, line 61: Should we keep:

> Be sure to remove the lines   that you are _not_ modifying, otherwise the contents of `current-values.yaml`   will override those in the default chart.

Or does it really not have any relevance?

setup/installation.md
jawnsy: One other small thing: I think command lines can be seen by anyone in /proc for multi-user systems, so putting the password in the command line means that it's potentially visible to everyone logged into the system. We can just document it though, too.

khorne3 (approved):

setup/installation.md
deansheather: No more relevance

ericpaulsen: Was that a change on the helm side?

setup/installation.md
sreya: I think doing it on the command line is fine. Most people aren't working on a multi-tenant box.

sreya: if it's easy or clean to show both ways so that we can avoid some rando opening a PR saying our docs leak secrets in a multi tenant system that would be ideal but either way this is pretty minor.

sreya (approved):

setup/installation.md
deansheather: We no longer get a full copy of the values file for the user, so they don't have to comment out anything

khorne3 (approved):

PR #402, opened 2021-06-11
chore: add CODER_DOCS_MAIN_BRANCH to build preview

BrunoQuaresma:


BrunoQuaresma:
I used `github.sha` because it works too 😄

.github/workflows/preview.yaml
```diff
@@ -53,6 +53,7 @@ jobs:
           VERCEL_ORG_ID: team_tGkWfhEGGelkkqUUm9nXq17r
           VERCEL_PROJECT_ID: QmZRucMRh3GFk1817ZgXjRVuw5fhTspHPHKct3JNQDEPGd
           VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
+          CODER_DOCS_MAIN_BRANCH: ${{ github.sha }}
```
jawnsy, line 56: In order to test that your change takes effect, you will also need to modify this line to check out your branch: https://github.com/cdr/docs/pull/402/files#diff-22da830e02ef5bed54ecad981722fd9c940aec797740c09927d4a904df0bccd6L20

You should be able to check the deployment log for the preview build and verify that Vercel emits the message the variable being overridden

BrunoQuaresma: Sorry, I did not understand this "In order to test that your change takes effect, you will also need to modify this line to check out your branch:".

jawnsy: This workflow works by cloning the monorepo's default branch (master), patching the docs submodule, and then deploying the monorepo and coder.com site to Vercel. Since you didn't modify the line that I referenced, it's cloning from master and you're not testing your changes.

In order to test your changes, you will need to change the reference that it checks out to be the branch in the monorepo that you made your changes on. Otherwise, this just shows that adding the environment variable doesn't break the existing build.

BrunoQuaresma: Ahhh ok, I gotcha. I will merge the new docs on master soon - any time today. So we can keep it as it is I think.

BrunoQuaresma: @jawnsy do you think we could merge this?

jawnsy (approved):

.github/workflows/preview.yaml
jawnsy: go for it :)

PR #404, opened 2021-06-11
add min glibc version for images

ericpaulsen:
adding min glibc version for the binaries we inject

[clubhouse story [ch8831]](https://app.clubhouse.io/coder/story/8831/add-an-image-requirements-doc)


images/requirements.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Image Requirements
+description: Learn about the minimum requirements for Coder images
+---
+
+Coder inserts assets into each workspace to enable code-server, JetBrains IDEs,
+and the Coder CLI. These assets are installed into the `/var/tmp/coder` directory
+of each workspace.
+
+The minimum GNU libc version supported for such binaries is 2.1.
```
jawnsy, line 10: Can this be part of the page where we describe how to make custom images? https://coder.com/docs/images/structure

I think we should avoid very small pages like this because it's more work to find maintain. This one also wouldn't be accessible from the menu since you didn't modify the JSON manifest

jawnsy (dismissed):

deansheather (changes requested):
This should have the other stuff from this page as well https://app.clubhouse.io/coder/story/8831/add-an-image-requirements-doc

khorne3 (approved):

images/structure.md
```diff
@@ -43,3 +43,29 @@ USER coder
...
+of each workspace.
+
+You do not need to include these assets in your own custom images. However, when
+building your own images, the following utilities are required:
+
+- [POSIX Utilities](https://pubs.opengroup.org/onlinepubs/9699919799/idx/utilities.html)
+- [GNU libc](https://www.gnu.org/software/libc/libc.html)
+  - The minimum GNU libc version supported for the Coder-inserted assets is `2.1`
+  - Alpine is not supported, since it uses musl libc
+- [GNU Core Utilities](https://en.wikipedia.org/wiki/List_of_GNU_Core_Utilities_commands)
```
ericpaulsen, line 60: I'm unsure if we are comfortable with referencing Wikipedia on our docs page. But this was the best listing of the utilities I could find.

images/structure.md
```diff
@@ -43,3 +43,29 @@ USER coder
...
+
+The following utilities are optional:
+
+- [ssh-agent](https://www.ssh.com/academy/ssh/agent) for automatically adding the
+Coder user's public SSH key to the agent
+- [systemd](https://systemd.io) for service supervision
+  - Only available with [CVMs](../workspaces/cvms)
+- [OpenSSH](https://www.openssh.com) server
+  - This can be run from either your `coder/configure` script or `systemd`
+- [gpg](https://www.gnupg.org) & [gpg-agent](https://www.gnupg.org/documentation/manuals/gnupg/Invoking-GPG_002dAGENT.html#Invoking-GPG_002dAGENT)
```
ericpaulsen, line 70: @deansheather in your clubhouse story, you marked the `gpg` utility as "future". let me know if I should add context here about a future supported capabilities.

images/structure.md
ericpaulsen, line 60: I'm unsure if we are comfortable with referencing Wikipedia on our docs page. But this was the best listing of the utilities I could find.

images/structure.md
```diff
@@ -43,3 +43,11 @@ USER coder
...
 tooling and configuration installed onto the base image and the new
 customizations added by the users.
+
+## Coder assets
+
+Coder inserts assets into each workspace to enable code-server, JetBrains IDEs,
+and the Coder CLI. These assets are installed into the `/var/tmp/coder` directory
+of each workspace.
+
+The minimum GNU libc version supported for such binaries is `2.1`.
```
jawnsy: Do you know if the symbol version reported by the objdump/ldd tools always matches the library/release version of glibc?

images/structure.md
```diff
@@ -43,3 +43,11 @@ USER coder
 Your users can then create descendant images. These contain all of the original
 tooling and configuration installed onto the base image and the new
 customizations added by the users.
+
+## Coder assets
+
+Coder inserts assets into each workspace to enable code-server, JetBrains IDEs,
```
jawnsy, line 49:
```suggestion
Coder injects static assets into each workspace, including code-server, JetBrains IDEs,
```

images/requirements.md
```diff
@@ -1,10 +0,0 @@
----
-title: Image Requirements
```
jawnsy: If you change this, don't you also need to modify the manifest?

images/structure.md
ericpaulsen: I don't know. Do you know where we can check the glibc library version?

images/requirements.md
ericpaulsen: we moved the contents of this PR to Images > Structure, so this isn't needed anymore.

images/structure.md
ericpaulsen: I think we were moving away from using the word "inject"? I'll let @khorne3 confirm.

jawnsy: If that's the case, then I suppose we should make similar changes to our in-product copy too

images/structure.md
```diff
@@ -43,3 +43,29 @@ USER coder
 Your users can then create descendant images. These contain all of the original
 tooling and configuration installed onto the base image and the new
 customizations added by the users.
+
+## Coder assets
+
+Coder inserts static assets into each workspace, including code-server, JetBrains
+IDEs, and the Coder CLI. These assets are installed into the `/var/tmp/coder` directory
```
deansheather, line 50: We don't inject jetbrains IDEs. I think you're confusing this with the code-server jetbrains plugin, which we don't need to document publicly.

images/structure.md
```diff
@@ -43,3 +43,29 @@ USER coder
...
+of each workspace.
+
+You do not need to include these assets in your own custom images. However, when
+building your own images, the following utilities are required:
```
deansheather, line 54:
```suggestion
building your own images, the following software is required:
```

images/structure.md
deansheather: ldd shows the version it was compiled with and the minimum it supports

images/structure.md
deansheather: maybe https://www.gnu.org/software/coreutils/

images/structure.md
deansheather: Probably don't include gpg until we make it a feature

images/structure.md
ericpaulsen: The prior wording was "assets that enable" the JetBrains IDEs, so I'll make the change to explicitly reference the Projector binary. Thanks for catching this.

jawnsy:
Removing my review request, I'm good with this if @deansheather is happy with it :smile:

images/structure.md
```diff
@@ -43,3 +43,36 @@ USER coder
...
+
+Coder inserts static assets into each workspace, including:
+
+- code-server
+- JetBrains Projector
+- Coder CLI
+
+These assets are installed into the `/var/tmp/coder` directory of each
+workspace. You do not need to include these statics assets in your custom
+images. However, the following software are **required** when you build custom
```
ericpaulsen, line 57: purely a grammatical question - is this supposed to be `the following software are` or `the following software is`?

deansheather: "is" I think, @khorne3

images/structure.md
```diff
@@ -43,3 +43,36 @@ USER coder
...
+
+- [ssh-agent](https://www.ssh.com/academy/ssh/agent) to automatically add the
+  Coder user's public SSH key to the agent
+- [systemd](https://systemd.io) for service supervision (this is only available
+  with [CVMs](../workspaces/cvms)
+- [OpenSSH](https://www.openssh.com) server
+  - You can run OpenSSH from either your `coder/configure` script or `systemd`
+- [gpg](https://www.gnupg.org) and
+  [gpg-agent](https://www.gnupg.org/documentation/manuals/gnupg/Invoking-GPG_002dAGENT.html#Invoking-GPG_002dAGENT)
+  for data encryption
```
deansheather, line 78:
```suggestion
```

deansheather (approved):

images/structure.md
khorne3: I actually Googled this because I didn't know -- software is both singular and plural, but I ended up choosing plural because the list has 2+ items

PR #410, opened 2021-06-15
chore: add CODER_RUNTIME var to docs

khorne3:


workspaces/variables.md
```diff
@@ -57,4 +57,8 @@ env | grep CODER_
         <td><code>CODER_WP_NAME</code></td>
         <td>The name of the workspace provider hosting the environment</td>
     </tr>
+    <tr>
+        <td><code>CODER_RUNTIME</code></td>
+        <td>Whether the workspace is a CVM or not</td>
```
khorne3, line 62: Maybe

> Whether the workspace is CVM-enabled or not

WDYT?

vapurrmaid: I think `is a` makes sense, since "containerized virtual machine" is a category, rather than something that can be turned on or off (this setting can't be changed after creation).

vapurrmaid (approved):

workspaces/variables.md
jawnsy: Sorry that I'm late to this, CODER_RUNTIME is set to the runtime name, currently either `kubernetes/default` or `kubernetes/sysbox` (for CVMs). The goal is that we would later support other runtimes, so this isn't a boolean flag: it is the name of the container runtime that we're using to start the workspace.

Code reference: https://github.com/cdr/m/blob/5cf9288a2b3e506dfa2b31098eaf730ce99a1a4c/product/coder/pkg/model/environments_stages.go#L283-L286

khorne3: @jawnsy Thanks, I'll update this in the release branch!

PR #411, opened 2021-06-15
Add info re: service banners

khorne3:


admin/appearance.md
```diff
@@ -6,20 +6,34 @@ description: Learn how to augment the dashboard appearance.
...
    in your **Header** and **Footer**.
 1. Click **Save Preferences** to save your changes.
 
 ![System appearance](../assets/system-banners.png)
 
+## Service banners
+
+The service banner allows you to display a message to all users of your Coder
+system. The user can dismiss the message at any time and Coder will not display
+a banner until you change the message.
```
vapurrmaid, line 28: To be explicit, the logic is:

- if the same message is disabled, then re-enabled, all users will see it
- if the message is updated, all users will see it

I don't know if we should clarify that in the wording, or this is satisfactory - doing a gut check, entirely up to you.

vapurrmaid (approved):

PR #412, opened 2021-06-15
Add docs re: browser security features

khorne3:


admin/security.md
```diff
@@ -0,0 +1,28 @@
...
+are available under **Manage** > **Admin** > **Infrastructure**.
+
+## HTTP Strict Transport Security
+
+Enabling **HTTP Strict Transport Security** sets [the
+`strict-transport-security`
+header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security)
+for Coder. If you're serving Coder over TLS, you can safely enable this header
+(most ingress options add this header). After you set this header, all users
+revisiting the site will use HTTPS, since clients may no longer access the site
```
jawnsy, line 16: I suggest removing the second clause ("since ...") because HSTS only has an effect for browsers that support it. HSTS is an instruction to the browser to always access the site over HTTPS, but very old browsers may not support it, and it is not generally guaranteed that clients won't access the site over HTTP, since it's a client-side feature.

admin/security.md
```diff
@@ -0,0 +1,28 @@
...
+Coder offers two browser security features that you can choose to enable. These
+are available under **Manage** > **Admin** > **Infrastructure**.
+
+## HTTP Strict Transport Security
+
+Enabling **HTTP Strict Transport Security** sets [the
+`strict-transport-security`
+header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security)
+for Coder. If you're serving Coder over TLS, you can safely enable this header
+(most ingress options add this header). After you set this header, all users
```
jawnsy, line 15:
```suggestion
(many ingress controllers automatically add this header for connections served over HTTPS). After you set this header, all users
```

admin/security.md
Emyrk: It's still a good idea to inform them though no? I mean _most_ users will be using browsers that support it.

Emyrk: We can remove the `since` if it reads better, but I think _most_ of our users will be using browsers that support HSTS.

Emyrk (approved):
I like the extra section!

admin/security.md
jawnsy: How about: "If you are serving Coder over HTTPS, we recommend enabling [HTTP Strict Transport Security], which requests that browsers and clients to use HTTPS for all connections."

It doesn't seem like mentioning ingress controllers are useful here?

Emyrk: That is fair.
The ingress thing was mentioned as it's covered by nginx by default. So anyone using that doesn't need this.

Idk, I'm ok with dropping the ingress mention, but this problem is usually resolved there.

jawnsy: My reasoning for dropping the ingress mention was that we're dropping the ingress controller requirement in a future release. It doesn't hurt to enable this if a proxy in front is also adding it, right?

Emyrk: It does not hurt, no

admin/security.md
```diff
@@ -11,10 +11,11 @@ are available under **Manage** > **Admin** > **Infrastructure**.
 Enabling **HTTP Strict Transport Security** sets [the
 `strict-transport-security`
 header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security)
-for Coder. If you're serving Coder over TLS, you can safely enable this header
-(most ingress options add this header). After you set this header, all users
-revisiting the site will use HTTPS, since clients may no longer access the site
-via HTTP.
+for Coder. If you are serving Coder over HTTPS, we recommend enabling HTTP
```
jawnsy, line 14: I think we can shorten this without losing meaning, I'll add a commit to this branch

admin/security.md
```diff
@@ -8,22 +8,23 @@ are available under **Manage** > **Admin** > **Infrastructure**.
...
 
 ## Secure Cookie
 
-Enabling **Secure Cookie** sets the `secure` property on all [session
-cookies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies), preventing
-browsers from sending sensitive cookies over unencrypted (HTTP) connections. We
-recommend enabling this setting if you are serving Coder over HTTPS.
+The **Secure Cookie** option controls the [`secure` property of cookies] that
+Coder issues. This prevents browsers from sending sensitive cookies, such as
+those containing credentials, over unencrypted (HTTP) connections.
```
jawnsy, line 25: oops, will add back the last sentence, seems useful

khorne3:
Thanks @jawnsy; all looks good!

vapurrmaid (approved):

jawnsy:
@khorne3 Since you're the original author of this, I'll let you merge when it makes sense to do so :)

PR #422, opened 2021-06-19
chore: use minimal permissions for workflows

jawnsy:
Restrict the GITHUB_TOKEN issued for builds to reads of respository
metadata and file contents.

### DIFF
--- a/.github/workflows/preview.yaml
+++ b/.github/workflows/preview.yaml
@@ -12,7 +12,7 @@ permissions:
   deployments: none
   issues: none
   packages: none
-  pull-requests: none
+  pull-requests: write
   repository-projects: none
   security-events: none
   statuses: none
--- a/.github/workflows/preview.yaml
+++ b/.github/workflows/preview.yaml
@@ -5,6 +5,18 @@ on:
     branches:
       - main
 
+permissions:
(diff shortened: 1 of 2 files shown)


vapurrmaid (changes requested):
Nice change! Unfortunately we're seeing this now:

https://github.com/cdr/docs/pull/422/checks?check_run_id=2865230382#step:10:15

Screenshot from 2021-06-21 10-03-02

jawnsy:
Ah good call, I was changing repo settings at the same time and broke that one, should be fixed now:

image

jawnsy:
Yeah, might need write access to pull_requests :thinking: will try that next

.github/workflows/preview.yaml
```diff
@@ -5,6 +5,18 @@ on:
...
       - main
 
+permissions:
+  actions: none
+  checks: none
+  contents: read
+  deployments: none
+  issues: none
+  packages: none
+  pull-requests: none
```
jawnsy, line 15:
```suggestion
  pull-requests: write
```

khorne3 (approved):

vapurrmaid (approved):

PR #423, opened 2021-06-21
chore: add subdirectories to /assets

khorne3:
- Create structure for /assets that mirror the /docs structure
- Move images into appropriate subdirectories
- Remove unused images


jawnsy:
Seems we have a problem in our build script:

```
fatal: 'images' could be both a local file and a tracking branch.
Please use -- (and optionally --no-guess) to disambiguate
```

We should fix this, but as a workaround, can you create a new branch name instead? You should be able to:

```
git checkout images
git checkout -b images-new
```

and create a new PR from that

khorne3:
Closed. See https://github.com/cdr/docs/pull/424

PR #424, opened 2021-06-21
chore: add subdirectories to /asset

khorne3:
- Create structure for /assets that mirror the /docs structure
- Move images into appropriate subdirectories
- Remove unused images


jawnsy (approved):
Seems good to me! I noticed that the images take a little while to load, so you get a blank page at first followed by an image appearing suddenly. I'm not sure if this was the previous behavior, though.

khorne3:
@jawnsy Thanks for the heads-up; I'm not sure if that's a result of moving things around, but I'll keep an eye on it and fix if it persists.

PR #431, opened 2021-06-23
chore: add 1.19.1 changelog

khorne3:


changelog/1.19.1.md
```diff
@@ -0,0 +1,21 @@
...
+---
+
+### Breaking changes ❗
+
+There are no breaking changes in 1.19.1.
+
+### Features ✨
+
+- infra: Added verbose logging around workspace provider startup process.
+- infra: Updated Sysbox container runtime version for CVMs
```
coadler, line 13: This is likely more of a bug, since this caused issues when using the latest version of `containerd`

khorne3: @coadler would the explanation then be `- infra: Updated Sysbox container runtime version for CVMs to fix issues when using the latest version of containerd`?

jawnsy: How about: `infra: Update Sysbox container runtime version to ensure that newer versions of containerd run properly` or something. I'm unsure about feature vs bug, I'm leaning toward feature because this is adding support for a new version of dockerd. The bug really is that our enterprise-base image updates to the latest version of Docker without giving us an opportunity to test

vapurrmaid: "new versions" - can we be more specific which versions we weren't supporting before but now are?

coadler: It wasn't just our images, the latest `ubuntu` images were broke too. I lean towards bug here since it should just be implicit we support newer minor versions of runc/containerd, but I don't feel super strongly either way. Both wordings make sense to me.

coadler: > "new versions" - can we be more specific which versions we weren't supporting before but now are?

It was just the latest release of containerd: https://github.com/containerd/containerd/releases/tag/v1.5.2

vapurrmaid (approved):

coadler (approved):

khorne3:
shipping this tomorrow (6/24)

deansheather (approved):

PR #449, opened 2021-06-29
chore: update installation step for SSL certs guide

khorne3:
Updating step 5 to reflect that cluster-issuer automatically issues certs using values from `helm install` for the dev URL and host secret names.

~I think this will need to be changed on all the guides and will do so once we've verified that the proposed updates are correct.~ This is the only doc that we explained certs this way.


khorne3:
CC: @mtm20176

ericpaulsen (approved):

vapurrmaid (approved):

PR #452, opened 2021-06-29
custom images: clarify building CVM images

bpmct:
Context:
- Docker is now [installed in the base image](https://github.com/cdr/enterprise-images/blob/main/images/base/Dockerfile.ubuntu#L29-L30), there is no need for my custom docker-sandbox image
- Specific images that leverage the sysbox runtime at build time, cannot be currently be built inside a Coder workspace.

### DIFF
--- a/images/writing.md
+++ b/images/writing.md
@@ -6,10 +6,12 @@ description: Learn how to write custom images for use with Coder.
 Custom images allow you to define workspaces that include the dependencies,
 scripts, and user preferences helpful for your project.
 
-> If you're unfamiliar with how to create, build, and push Docker Images, please
-> review [this tutorial by
-> Docker](http://blog.shippable.com/build-a-docker-image-and-push-it-to-docker-hub)
-> before proceeding.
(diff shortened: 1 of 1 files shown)


ericpaulsen (approved):

khorne3 (approved):

images/writing.md
```diff
@@ -7,8 +7,8 @@ Custom images allow you to define workspaces that include the dependencies,
 scripts, and user preferences helpful for your project.
 
 > If you're unfamiliar with how to create, build, and push Docker Images, please
-> review
-> [this tutorial by Docker](http://blog.shippable.com/build-a-docker-image-and-push-it-to-docker-hub)
+> review [this tutorial by
```
jawnsy: the label implies that Docker Inc wrote this, but the post is by a different company?

The link also appears to be dead

khorne3: Updating to link to docs; unfortunately, the blog is having some CDN issues (I think) but we don't want to rely on them despite this

PR #454, opened 2021-06-30
Add production workflow

BrunoQuaresma:
This workflow is based on https://github.com/cdr/m/blob/master/.github/workflows/codercom.yaml#L83 and https://github.com/cdr/docs/blob/main/.github/workflows/preview.yaml#L21. I removed the DatoCMS update since this is not related to it at all.

PS: Idk how would be the best way to test this action though.

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -1,4 +1,4 @@
-name: Preview
+name: Production
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


jawnsy:
There isn't really a great way to test this, you could manually add a `pull_request: branches: main` build to trigger the production build for your PR, but otherwise we can just merge and fix if it's broken

jawnsy (approved):

PR #455, opened 2021-06-30
Add a GCR page to registries

mterhar:
ch-14509


mterhar:
Threw this together in the web editor so it's probably ugly. Just wanted to get the ball rolling on it. This can also be a guide or a "cloud registries" page where we have sections for GCR, Azure's registry, dockerhub, etc. Some will be easier than others. Also explaining why ECR is troublesome and providing Ben's script that updates the auth frequently may help some users though.

khorne3 (approved):

PR #461, opened 2021-07-01
Airgapped network instructions cert application command is wrong

mterhar:
The cert application command needs to be a helm upgrade command rather than a kubectl get secret command.


khorne3 (approved):

khorne3:
Would prefer to leave as `console` for consistency throughout the site

PR #465, opened 2021-07-06
update letsencrypt server URL to prod

ericpaulsen:
making this change to prevent untrusted certs from being used, and providing a sub-optimal TLS config experience for users.

### DIFF
--- a/guides/ssl-certificates/route53.md
+++ b/guides/ssl-certificates/route53.md
@@ -104,7 +104,7 @@ To make sure that your `clusterIssuer` can change your DNS settings,
        preferredChain: ""
        privateKeySecretRef:
          name: example-issuer-account-key
-       server: https://acme-staging-v02.api.letsencrypt.org/directory
+       server: https://acme-v02.api.letsencrypt.org/directory
        solvers:
          - dns01:
              route53:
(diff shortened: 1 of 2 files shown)


IAmHughes (approved):
thanks for the quick PR

kylecarbs (approved):
Do we use staging anywhere else in the docs?

ericpaulsen:
> Do we use staging anywhere else in the docs?

Not that I'm aware of. The other guides actually had the production URL set, so this change will make the guides consistent. 

Side note: I'm going to push another commit to fix an issue with the `devurls.host` value we're setting in the `helm upgrade` command.

khorne3 (approved):

guides/ssl-certificates/route53.md
```diff
@@ -104,7 +104,7 @@ To make sure that your `clusterIssuer` can change your DNS settings,
        preferredChain: ""
        privateKeySecretRef:
          name: example-issuer-account-key
-       server: https://acme-staging-v02.api.letsencrypt.org/directory
+       server: https://acme-v02.api.letsencrypt.org/directory
```
jawnsy, line 107: Should we provide examples for both? I imagine that someone would want to test with staging first, and then switch to production once they confirm it works, because I think the rate limits are different between staging and prod?

ericpaulsen: Correct - the rate limits are different. [See here](source /tmp/pure_installer.fish; and install_pure). I'm in favor of including an example for each, so users are aware of both options and can decide for themselves.

PR #474, opened 2021-07-09
chore: add notes on using AWS LB

khorne3:


bpmct (commented):
Has this been tested/do we have an EKS cluster we can test this with?

I do not use AWS/EKS commonly but can volunteer to test if the answers above are no. 

Also including @mterhar in the review since it looks like he had the initial notes for this (I checked Clubhouse)

ericpaulsen:
> Has this been tested/do we have an EKS cluster we can test this with?
> 
> I do not use AWS/EKS commonly but can volunteer to test if the answers above are no.
> 
> Also including @mterhar in the review since it looks like he had the initial notes for this (I checked Clubhouse)

I can spin up a cluster on EKS and test this out.

khorne3:
@bpmct do you still have time to tackle this? If not, just let me know and I'll take care of it. Thanks!

bpmct:
I can do this after Wednesday, if we'd like to merge it before then, that would be great :)

@ericpaulsen @mterhar if this has been tested already, we can go ahead and merge. I just don't want to push untested docs and I don't have a lot of context behind this PR.

ericpaulsen:
@bpmct I was unsuccessful in my attempts last week to get the LB routing traffic properly. I didn't do much troubleshooting after the first couple of deploys, since I got caught up with other tasks. That said, there are certain values you'll need to set/play with:

- access URLs: Mike & I were able to get through some errors by setting both the `cemanager` & `envproxy` access URLs
- TLS: disabling TLS got things working better than keeping it enabled, for reasons I do not know. something to keep in mind
- make sure to set the ALB annotations under `ingress.service.annotations`, as this will modify the `LoadBalancer` service used by the Coder ingress

bpmct:
Suggest closing this issue or revisiting solutions. Mike reported some issues with running Coder + an AWS LoadBalancer

khorne3:
@bpmct Closing this, because I don't think we really have a working solution to document at the moment.

PR #476, opened 2021-07-09
chore: expand instructions on creating a configure file

khorne3:
Minor fixed based on customer confusion -- we had this info super buried


bpmct (approved):
good find. this has happened to me a few occasions

PR #486, opened 2021-07-12
chore: document RStudio usage

khorne3:
**Note:** Hold for 1.21 release.


workspaces/editors.md
```diff
@@ -178,3 +179,36 @@ participants. If they do, click **Accept** to proceed.
...
+1. Create an image that includes RStudio (see
+   [rocker/rstudio](https://hub.docker.com/r/rocker/rstudio) for an example).
+
+1. [Create a workspace](getting-started.md#2-create-a-workspace) using the image
+   you created in the previous step.
+
+1. Because RStudio uses an authentication model that matches to your Linux user
+   with a home directory, whereas your image likely uses the `coder` user, we
+   recommend making a new user for your RStudio projects. To do so, go to
+   **Applications** > **Terminal** in Coder and run:
```
vapurrmaid, line 197: I'd adjust the wording here to be more around user with home directory and known password. Something along these lines:

1. (Optional) RStudio requires authentication using a Linux user and stores all files in their home directory. If from your Docker image you're using the root user, `coder` user, or a user that either does not have a home directory or known password, then we recommend creating a new user for your RStudio projects. To do so, go to...

```console
useradd --create-home <username>
passwd <username>
```

workspaces/editors.md
```diff
@@ -178,3 +179,36 @@ participants. If they do, click **Accept** to proceed.
...
+
+1. At this point, you can go to **Applications** to launch RStudio:
+
+   ![Applications with RStudio launcher](../assets/workspaces/rstudio.png)
+
+   Sign in using your RStudio user credentials (RStudio auto-populates
+   the fields your Coder credentials, so you'll need to change these to the ones
+   you just created).
+
+   > All RStudio user data is store in the `home` directory associated with this
```
vapurrmaid, line 212:
```suggestion
   > All RStudio user data is stored in the `home` directory associated with this
```

vapurrmaid (commented):
Lookin' great!

workspaces/editors.md
```diff
@@ -178,3 +179,37 @@ participants. If they do, click **Accept** to proceed.
...
+1. At this point, you can go to **Applications** to launch RStudio:
+
+   ![Applications with RStudio launcher](../assets/workspaces/rstudio.png)
+
+   Sign in using your RStudio user credentials (RStudio auto-populates
+   the fields your Coder credentials, so you'll need to change these to the ones
+   you just created).
+
+   > All RStudio user data is stored in the `home` directory associated with this
+   > user, not your Coder user (that is, you'll find it in `/home/myuser` where
```
vapurrmaid, line 214: I think the concept of "your Coder user" isn't obviously clear: the reason we and many are set to a user `coder` is because we in our images (https://github.com/cdr/enterprise-images) set the `USER` to `coder`:

https://github.com/cdr/enterprise-images/blob/11926804262ead6c2ac41061488ddcd043868216/images/node/Dockerfile.ubuntu#L16

There's nothing inherently tied to the user of `coder` otherwise; if I homebrew an image like so:

```dockerfile
FROM node:latest
USER myuser
```

Then my home directory will be `/home/myuser`.

As a result - I'm not entirely sure of the best wording, but I'd want to refrain from "your Coder user" to avoid confusion in cases whereby folks are using a custom non-Coder image.

Maybe something like:

> All RStudio data is stored in the home directory associated with this user, which may be different than the user on workspace startup (that is, you'll find the data in `/home/myuser` where `myuser` is your RStudio username).

deansheather:
I don't know anything about the RStudio support, you may want to add the person responsible for adding the feature instead (not sure who that is though)

bpmct (changes requested):
I can help with these since I need RStudio screenshots for the 1.21 blog.

workspaces/editors.md
```diff
@@ -178,3 +179,38 @@ participants. If they do, click **Accept** to proceed.
...
+1. **Optional.** RStudio requires authentication with a Linux user and stores
+   all files in their home directory. If, from your Docker image, you're using the
+   root user, `coder`, or a user that doesn't have a home directory
+   or known password, then we recommend creating a new user for your RStudio
+   projects. To do so, run:
+
+   ```console
+   useradd --create-home myuser     # replace myuser with your username
+   passwd myuser     # provide and confirm your chosen password when prompted
+   ```
```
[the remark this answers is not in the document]
khorne3: @scsmithr do you have thoughts on which option we should recommend to users?

workspaces/editors.md
```diff
@@ -178,3 +179,38 @@ participants. If they do, click **Accept** to proceed.
 
 At this point, you'll be able to share your project and work with your partner
 in real-time.
+
+## RStudio
```
scsmithr, line 183: I think we should be explicit, and say we support RStudio Server. RStudio and RStudio Server both seem to be functionally equivalent, but I don't have enough experience with them to say that they're interchangeable.

workspaces/editors.md
```diff
@@ -178,3 +179,38 @@ participants. If they do, click **Accept** to proceed.
...
+
+Coder supports [RStudio](rstudio.com). To create a workspace that lets you use
+RStudio:
+
+1. Create an image that includes RStudio (see
+   [rocker/rstudio](https://hub.docker.com/r/rocker/rstudio) for an example).
```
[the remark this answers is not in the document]
scsmithr: I'm sort of in agreement. `rocker/rstudio` _seems_ to be the image that most RStudio people use, but it does a lot of stuff that isn't necessary for just running rstudio.

Also fun fact, `rocker/rstudio` will error on startup if a password environment variable isn't set (https://github.com/rocker-org/rocker-versioned2/blob/3456299fbbd1190a0aefab7ad6e0a3239932a0b4/scripts/userconf.sh#L26-L35). However, when this image is used in Coder, it doesn't error because the entrypoint for the image is overwritten, meaning that this script doesn't get ran.

workspaces/editors.md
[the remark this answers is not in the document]
vapurrmaid: @bpmct - you're correct, I incorrectly thought `/home/myuser` would persist. We should fix this suggestion. Given that RStudio data would be removed too, I think we need to make sure that the USER in the Dockerfile is compatible with RStudio.

workspaces/editors.md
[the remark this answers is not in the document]
vapurrmaid: We should - and it will have to have a known `passwd` for the user

scsmithr:
Also, we should probably add a known issues section for https://github.com/cdr/code-server-rstudio/issues/8. Very minor (a refresh fixes it), but it is a bug.

workspaces/editors.md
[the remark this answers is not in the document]
scsmithr: We should make sure these requirements are listed as well:
- `pgrep` and `rserver` must be in the shell's PATH.
- `rserver` must be configured to start on port 8787 (the default).

admin/workspace-management/installing-jetbrains.md
```diff
@@ -24,6 +24,9 @@ PATH:
 - PhpStorm
 - PyCharm
 - Rider
+- RStudio (see
+  [rocker-versioned](https://github.com/rocker-org/rocker-versioned/tree/master/rstudio)
+  for sample images installing RStudio)
```
vapurrmaid, line 29: @scsmithr @bpmct - should we remove this?

workspaces/editors.md
```diff
@@ -168,13 +169,91 @@ If you've received a link to join a Code With Me session as a participant:
...
+1. [Create a workspace](getting-started.md#2-create-a-workspace) using the image
+   you created in the previous step.
+
+1. At this point, you can go to **Applications** to launch RStudio.
+
+   ![Applications with RStudio launcher](../assets/workspaces/rstudio.png)
+
+   Sign in using your RStudio user credentials (RStudio auto-populates the
+   fields your Coder credentials, so you'll need to change these to the ones you
+   just created).
```
vapurrmaid, line 251:
```suggestion
   Sign in using your linux username and password (as defined in your image)
```

workspaces/editors.md
```diff
@@ -168,13 +169,91 @@ If you've received a link to join a Code With Me session as a participant:
...
+
+   > RStudio may take a few additional seconds to start launch after the
+   > workspace is built.
+   >
+   > All RStudio data is stored in the home directory associated with the user
+   > you sign in as, which may be different than the user on workspace startup
+   > (that is, you'll find the data in `/home/myuser` where myuser is your
+   > RStudio username).
```
vapurrmaid, line 259:
```suggestion
   > All RStudio data is stored in the home directory associated with the user
   > you sign in as
```

vapurrmaid (commented):
Just some cleanups we can make now that we have the Dockerfile example

workspaces/editors.md
```diff
@@ -185,31 +185,73 @@ in real-time.
...
+   # create coder user
+   RUN useradd coder \
+   --create-home \
+   --shell=/bin/bash \
+   --uid=1000 \
+   --user-group && \
+   echo "coder ALL=(ALL) NOPASSWD:ALL" >>/etc/sudoers.d/nopasswd
+
+   # assign password "rstudio" to coder user.
+   RUN chown -R coder:coder /var/lib/rstudio-server RUN echo
```
jawnsy, line 227: looks like some of these lines weren't wrapped correctly?

vapurrmaid (approved):
Everything looks good to me with @scsmithr 's suggested change ✔️

bpmct (approved):

PR #490, opened 2021-07-14
Update references to ECDSA.

johnstcn:
* Update 1.21 security release notes with details of ECDSA -> ED25519 algorithm change
* Update existing references to ECDSA where applicable.
* Drive-by: fix typo

### DIFF
--- a/admin/git.md
+++ b/admin/git.md
@@ -23,7 +23,7 @@ Instead, you can use Visual Studio Code with git, the command-line tool, and we
 expect that this combination will work with most hosting software or services.
 However, Coder doesn't test these and cannot provide recommendations or support.
 
-**Note** The public key provided by Coder uses the ED25519 key algorithm. Ensure
+**Note** The public key provided by Coder uses the Ed25519 key algorithm. Ensure
 your git provider supports this.
 
 ![Configure Git Integration](../assets/admin/git-admin.png)
--- a/admin/git.md
+++ b/admin/git.md
@@ -23,7 +23,7 @@ Instead, you can use Visual Studio Code with git, the command-line tool, and we
 expect that this combination will work with most hosting software or services.
 However, Coder doesn't test these and cannot provide recommendations or support.
 
-**Note** The public key provided by Coder uses the ECDSA key algorithm. Ensure
+**Note** The public key provided by Coder uses the ED25519 key algorithm. Ensure
 your git provider supports this.
 
 ![Configure Git Integration](../assets/admin/git-admin.png)
--- a/changelog/1.21.0.md
+++ b/changelog/1.21.0.md
@@ -17,4 +17,5 @@ There are no bug fixes in 1.21.0.
 
 ### Security updates 🔐
 
-There are no security updates in 1.21.0.
+- ssh: New keys are generated using the [ED25519 digital signature algorithm](https://datatracker.ietf.org/doc/html/rfc8709)
+instead of the ECDSA digital signature algorithm.
--- a/changelog/1.21.0.md
+++ b/changelog/1.21.0.md
@@ -17,5 +17,6 @@ There are no bug fixes in 1.21.0.
 
 ### Security updates 🔐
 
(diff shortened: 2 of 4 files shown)


admin/git.md
```diff
@@ -23,7 +23,7 @@ Instead, you can use Visual Studio Code with git, the command-line tool, and we
 expect that this combination will work with most hosting software or services.
 However, Coder doesn't test these and cannot provide recommendations or support.
 
-**Note** The public key provided by Coder uses the ECDSA key algorithm. Ensure
+**Note** The public key provided by Coder uses the ED25519 key algorithm. Ensure
```
jawnsy, line 26: I'm not an expert at this kind of thing, but [according to the Wikipedia page](https://en.wikipedia.org/wiki/EdDSA#Ed25519), it should be stylized "Ed25519"

```suggestion
**Note** The public key provided by Coder uses the Ed25519 key algorithm. Ensure
```

changelog/1.21.0.md
```diff
@@ -17,4 +17,5 @@ There are no bug fixes in 1.21.0.
 
 ### Security updates 🔐
 
-There are no security updates in 1.21.0.
+- ssh: New keys are generated using the [ED25519 digital signature algorithm](https://datatracker.ietf.org/doc/html/rfc8709)
```
jawnsy, line 20: Didn't this also cause some sort of breakage with existing ECDSA keys? That's what I noticed in our dogfood environment, though I'm unsure how to replicate

workspaces/preferences.md
```diff
@@ -40,9 +40,10 @@ need to access from your workspace.
 If necessary, you can regenerate your key. Be sure to provide your updated key
 to all of the services you use. Otherwise, they won't work.
 
-> The SSH public key uses the ECDSA key algorithm instead of the more commonly
-> used RSA key. This provides you with a smaller, more performant key with the
-> same level of security as RSA.
+> The SSH public keys use the [ED25519 key algorithm](https://datatracker.ietf.org/doc/html/rfc8709)
+> instead of the more commonly used RSA algorithm.
+> This provides you with a smaller, more performant key with the same level of
```
jawnsy, line 45: I don't know the history of this statement but seems like we should be careful about making claims about "same level of security" unless we have a citation from a relevant authority. It might be better to use a phrase similar to Wikipedia's explanation of EdDSA - something like: "The SSH keys that Coder provisions use the [Ed25519 key algorithm], which was designed to be faster than predecessors such as RSA, with smaller key sizes, without sacrificing security"

@deansheather Do you have thoughts on this?

jawnsy (approved):
LGTM! Also check with @khorne3 to review/edit before merging :)

changelog/1.21.0.md
johnstcn: I wasn't able to replicate the issue you faced, but it sounds like you may have regenerated the SSH key while the workspace was still running, which would result in the new key being updated on GitHub and in `coderd` while the workspace still contained the old key. Rebuilding the workspace would bring everything back in sync again, but this would happen regardless of the key algorithm used.

workspaces/preferences.md
deansheather: GitLab says in their docs that ed25519 is better than RSA so we should just go with that and cite the same source https://docs.gitlab.com/ee/ssh/index.html#ed25519-ssh-keys

deansheather (approved):

workspaces/preferences.md
jawnsy: Thanks, that's a great reference. The language in that document is definitely weaker than we had ("Available documentation suggests that ED25519 is more secure than RSA.") which I think is the right approach to take with such things

changelog/1.21.0.md
jawnsy: Makes sense, thanks for double-checking!

workspaces/preferences.md
johnstcn: Done, thanks for the reference @deansheather !

khorne3 (approved):

PR #491, opened 2021-07-14
chore: add workspace diagrams

khorne3:


workspaces/index.md
```diff
@@ -20,11 +20,15 @@ icon:
 description: "Learn about development workspaces in Coder."
 ---
 
-A workspace is your remote development environment. All workspaces are based on
-a shared [image](../images/index.md) and therefore reproducible.
+A workspace is your remote development environment.
 
 ![Workspaces](../assets/workspaces/workspaces.svg)
 
+All workspaces are based on a shared [image](../images/index.md) and therefore
```
jawnsy, line 27:
```suggestion
Coder creates workspaces using a [shared container image](../images/index.md), which improves their reproducibility.
```

The phrase "reproducible" means different things to different people, so we should either elaborate on what we mean by reproducible here, or acknowledge that Coder increases reproducibility but cannot guarantee it (many subtle things make builds non-reproducible, and the [Reproducible Builds project](https://reproducible-builds.org/) describes more of the background)

workspaces/index.md
```diff
@@ -20,11 +20,15 @@ icon:
...
+reproducible.
+
+![The anatomy of a workspace](../assets/workspaces/workspace-anatomy.png)
```
jawnsy, line 30: @jmilling33 this image has a typo, `coder.taml`

workspaces/getting-started.md
```diff
@@ -3,6 +3,11 @@ title: "Getting started"
 description: "Learn how to develop in a workspace."
 ---
 
+This article walks you through creating a new workspace using a pre-defined
+image.
+
+![Relationship between images and workspaces](../assets/workspaces/lifecycles.png)
```
jawnsy, line 9: This image looks pretty complex and I'm not sure what we're trying to get at here... @mterhar I think this was originally a sketch you put together, right? Can we add some explanatory text that describes what we're trying to demonstrate with this image?

workspaces/index.md
jmilling33: Just fixed!
Copy of Anatomy of a Coder WorkspaceFinal (5)

jawnsy: @jmilling33 Thanks! For the examples, I think our style is to use "e.g.," (with the comma) rather than "ex."

I don't think we need the parentheses in "Cloud IDEs"

jmilling33: Gotcha. Thanks for the catch. 
Copy of Anatomy of a Coder WorkspaceFinal (6)

workspaces/getting-started.md
khorne3: I don't know if this is the greatest space for it -- I just wanted to try it out and see what y'all think. I don't know if we would be better off turning this visual into a guide on how Coder works under-the-hood, so to speak?

bpmct: This was originally a diagram I created for a blog post. Will defer to y'all but I think this would be well suited for the workspaces/index and when referring to workspace templates :)

bpmct (approved):

PR #499, opened 2021-07-20
add `value` to configure script examples

ericpaulsen:
configure script examples were missing the `value` piece. user sent in a zendesk ticket notifying us of this.


Emyrk (approved):
Good catch

khorne3 (approved):

PR #503, opened 2021-07-21
fix 1.19.1 release date

ericpaulsen:
changelog mentioned that `1.19.1` was released on 06/23/2021, which is _after_ the `1.20` release date. I assume the correct date is 05/23/2021, but if someone could double check, that would be great.

### DIFF
--- a/changelog/1.19.1.md
+++ b/changelog/1.19.1.md
@@ -1,6 +1,6 @@
 ---
 title: "1.19.1"
-description: "Released on 06/23/2021"
+description: "Released on 05/23/2021"
 ---
 
 ### Breaking changes ❗


jawnsy (changes requested):
I think the original release date is correct. You can verify this by checking the tags in the monorepo: https://github.com/cdr/m/releases/tag/1.19.1 and https://github.com/cdr/m/releases/tag/1.20.0

The reason for this is that we maintain multiple versions in parallel, since we backport critical fixes - such as security fixes - and create patch releases on prior versions, on a case-by-case basis.

ericpaulsen:
Ah good to know - I'll be sure to check the monorepo tags next time. Thank you for the clarification. I'll close this PR out.

PR #504, opened 2021-07-22
chore: rename SSL to TLS

khorne3:


jawnsy:
Thanks for the quick work on this! :clap:

We'll also need to update our redirects to make sure the old links continue working, and update any in-product links to refer to the new locations (if any)

[ch15295]

jawnsy (approved):

PR #510, opened 2021-07-26
recommend default formatter and auto-format on save

bpmct:
This repository recommends the `davidanson.vscode-markdownlint` extension and prompts the user to install it when they first clone the docs repo. However, the formatter will do nothing unless people set it to their default formatter and intermittently use the "format" command in the command palette or a keyboard shortcut.

This proposed PR will (for all users editing the docs with VS Code):
- set `davidanson.vscode-markdownlint` as the default formatter for this repo
- auto format (with this `davidanson.vscode-markdownlint`) whenever a file is saved

This should also lead to less linting issues from new contributors

Alternatives:
- document how someone can do this themselves on the README

I use the "prettier" extension for other projects and, while it always passes lint tests, it will try to "re-format" some docs. I found the issue was my default formatter wasn't set (or was reset) for this repo.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,6 +1,7 @@
 {
   "recommendations": [
     "davidanson.vscode-markdownlint",
-    "streetsidesoftware.code-spell-checker"
+    "streetsidesoftware.code-spell-checker",
+    "stkb.rewrap"
   ]
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,5 @@
 {
-  "editor.rulers": [80]
+  "editor.rulers": [80],
+  "editor.defaultFormatter": "DavidAnson.vscode-markdownlint",
+  "editor.formatOnSave": true
 }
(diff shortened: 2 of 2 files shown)


bpmct:
@khorne3 does markdownlint automatically fix your line errors for you? Prettier did this for me but with markdownlint I am finding that I have to manually decide when to break a line which feels a bit awkward and clunky. 

Screen Shot 2021-07-26 at 9 52 37 AM

khorne3:
@bpmct I'm using Rewrap to automatically add linebreaks because mdlint doesn't

bpmct:
> @bpmct I'm using Rewrap to automatically add linebreaks because mdlint doesn't

Is it [stkb.rewrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap)? Are there any VS Code settings you use to "rewrap" on save we can add here?

khorne3:
Yup, that's the one. I checked my JSON file and all I have is `"rewrap.autoWrap.enabled": true,` (top level)

vapurrmaid (approved):

khorne3 (approved):

PR #516, opened 2021-07-27
chore: fix wording in changelog

khorne3:


khorne3:
@bpmct @CoderErik Can I get y'all's thoughts on whether the update to the satellites line is clearer please?

CoderErik (approved):
Seems clearer to me. I suggested one little tweak as well.

PR #517, opened 2021-07-27
Specify single tenant AD

mterhar:
@IAmHughes if there are other take-always, toss them in here.


IAmHughes:
@mterhar I think the fix @sreya and @scsmithr are working on with @kylecarbs is going to be to allow multi-tenants. We can do this doc merge, just need to make sure we update the docs again when we fix it to support multi-tenant 👍

khorne3 (approved):

PR #518, opened 2021-07-28
fix TLS link in devurls page

bpmct:


mtm20176 (commented):
looks good, thanks for fixing this

mtm20176 (approved):

PR #520, opened 2021-07-28
remove instructions around helm uninstall

f0ssel:
There's a bug in this flow that is hard to recover from, removing for now so no one fall into it and will add back new docs ASAP


Emyrk (approved):
Should get the resolved docs in asap

PR #523, opened 2021-07-29
chore: expand upgrade instructions for 1.21

khorne3:
[Preview](https://codercom-4z3kbvwnq-codercom.vercel.app/docs/coder/v1.21/setup/updating)

### DIFF
--- a/setup/updating.md
+++ b/setup/updating.md
@@ -68,7 +68,9 @@ To update Coder, follow these steps:
 
    ```
 
-1. [Update your Helm chart](../guides/admin/helm-charts.md) as follows:
+1. [Update your Helm chart](../guides/admin/helm-charts.md) via the steps below.
+Refer to the `1.21` [helm chart](https://github.com/cdr/enterprise-helm/blob/main/values.yaml)
+when making such changes:
 
    Rename all instances of `cemanager` to `coderd`.
 
@@ -82,6 +84,8 @@ To update Coder, follow these steps:
    controller, set `coderd.serviceNext` to `true`. (See our
    [TLS certificates](../guides/tls-certificates/index.md) for more information
    on how to configure Coder to issue and use certificates.)
+
+1. Migrate the following ports. This may need to be done within your cloud provider:
 
    Move TCP port `8080` to `80`.
 
--- a/setup/updating.md
+++ b/setup/updating.md
@@ -68,21 +68,36 @@ To update Coder, follow these steps:
 
    ```
 
-1. Pull your updated Helm chart and
-   [make the the following changes](../guides/admin/helm-charts.md):
+1. [Update your Helm chart](../guides/admin/helm-charts.md) as follows:
 
-   a. Rename all instances of `cemanager` to `coderd`. b. Remove all
-   `dashboard.*` values. Route any traffic to `coderd`. c. Remove all
-   `envproxy.*` values. Route any traffice to `coderd`. d. Redirect all traffic
-   from your custom ingress to `coderd`. e. `coderd` now serves TLS
-   certificates; to upgrade the NGINX ingress controller, set
-   `coderd.serviceNext` to `true`. f. Move TCP port `8080` to `80`. g. Move TCP
-   port `8443` to `443`.
+   Rename all instances of `cemanager` to `coderd`.
 
-1. Upgrade workspaces to use Coder's new Networking v2 functionality. If all of
-   your workspaces are local, go to Manage > Providers. Find your provider,
-   click the **vertical ellipses** to its right, and click **Edit**. Toggle
-   **Networking v2** to enable.
+   Remove all `dashboard.*` values. Route any traffic to `coderd`.
+
+   Remove all `envproxy.*` values. Route any traffic to `coderd`.
+
+   Redirect all traffic from your custom ingress to `coderd`.
+
+   `coderd` now serves TLS certificates; to upgrade the NGINX ingress
+   controller, set `coderd.serviceNext` to `true`. (See our
+   [TLS certificates](../guides/tls-certificates/index.md) for more information
+   on how to configure Coder to issue and use certificates.)
+
+   Move TCP port `8080` to `80`.
+
+   Move TCP port `8443` to `443`.
+
+1. Update Coder with your new Helm chart values:
+
+   ```console
+   helm upgrade coder coder/coder -n coder --version=<VERSION> --values current-values.yaml
+   ```
+
+1. Upgrade workspaces to use Coder's new Networking v2 functionality.
+
+   If all of your workspaces are local, go to Manage > Providers. Find your
+   provider, click the **vertical ellipses** to its right, and click **Edit**.
+   Toggle **Networking v2** to enable.
 
    Alternatively, if your workspaces are distributed, migrate the providers to
    [satellites](../admin/satellites/index.md).
--- a/setup/updating.md
+++ b/setup/updating.md
@@ -52,7 +52,7 @@ This guide will show you how to update your Coder deployment.
    helm get values --namespace coder coder > current-values.yaml
    ```
 
-   > Make sure that your values only contain the changes you want (i.e., if you
+   > Make sure that your values only contain the changes you want (e.g., if you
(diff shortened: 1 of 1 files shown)


ericpaulsen:
I'm thinking we can dogfood this by upgrading `demo` to `1.12` - thoughts? @mterhar @bpmct

setup/updating.md
```diff
@@ -65,8 +65,43 @@ To update Coder, follow these steps:
...
+   on how to configure Coder to issue and use certificates.)
+
+   Move TCP port `8080` to `80`.
+
+   Move TCP port `8443` to `443`.
+
+1. Update Coder with your new Helm chart values:
+
+   ```console
+   helm upgrade coder coder/coder -n coder --version=<VERSION> --values current-values.yaml
```
ericpaulsen, line 93: We should clarify that updating the helm chart is only need when upgrading from `1.20` to `1.21`.

khorne3:
@ericpaulsen if you could that would be great; I tried this on my cluster but a second set of eyes would be v helpful just in case

ericpaulsen:
@khorne3 - I pushed a commit adding some context. Other than that, we are good to merge.

bpmct:
what is the context here? I thought that 1.20 -> 1.21 was non-breaking?

ericpaulsen:
it's not that installs have been breaking (to my knowledge). but rather, there are various helm values that have been deprecated (e.g. `dashboard`, `timescale`, `envproxy`, etc.), so this addition helps customers correctly update their helm values.

bpmct:
> it's not that installs have been breaking (to my knowledge). but rather, there are various helm values that have been deprecated (e.g. `dashboard`, `timescale`, `envproxy`, etc.), so this addition helps customers correctly update their helm values.

Makes sense to me

setup/updating.md
```diff
@@ -65,8 +65,47 @@ To update Coder, follow these steps:
...
+Refer to the `1.21` [helm chart](https://github.com/cdr/enterprise-helm/blob/main/values.yaml)
+when making such changes:
+
+   Rename all instances of `cemanager` to `coderd`.
+
+   Remove all `dashboard.*` values. Route any traffic to `coderd`.
+
+   Remove all `envproxy.*` values. Route any traffic to `coderd`.
+
+   Redirect all traffic from your custom ingress to `coderd`.
```
bpmct, line 81: It seems like sentence 1, 2 are covered by sentence 3. coderd inherits some helm values that used to be under dashboard.* and envproxy.* but not all.

setup/updating.md
```diff
@@ -65,8 +65,47 @@ To update Coder, follow these steps:
...
+1. [Update your Helm chart](../guides/admin/helm-charts.md) via the steps below.
+Refer to the `1.21` [helm chart](https://github.com/cdr/enterprise-helm/blob/main/values.yaml)
+when making such changes:
+
+   Rename all instances of `cemanager` to `coderd`.
+
+   Remove all `dashboard.*` values. Route any traffic to `coderd`.
+
+   Remove all `envproxy.*` values. Route any traffic to `coderd`.
+
```
bpmct, line 80: SSH and cemanager.accessurl are also no longer managed by the helm chart.

SSH is done in the Manage > Providers page. The old value used to be [ssh.enable](https://github.com/cdr/enterprise-helm/commit/2cf1185503b7cf06f2097269a95eb460348aa657#diff-3c79bd4a49240f41e502e8e5ff347991776222adb25b3c5fe6386ac94c31e116L66)

Access URL is now just in Manage > Admin under the infrastructure page.

setup/updating.md
```diff
@@ -65,8 +65,47 @@ To update Coder, follow these steps:
...
+   Redirect all traffic from your custom ingress to `coderd`.
+
+   `coderd` now serves TLS certificates; to upgrade the NGINX ingress
+   controller, set `coderd.serviceNext` to `true`. (See our
+   [TLS certificates](../guides/tls-certificates/index.md) for more information
+   on how to configure Coder to issue and use certificates.)
+
+1. Migrate the following ports. This may need to be done within your cloud provider:
```
bpmct, line 88: This is for custom ingresses right? The LoadBalancer should still pick up the service fine.

setup/updating.md
```diff
@@ -65,8 +65,47 @@ To update Coder, follow these steps:
...
+
+1. Update Coder with your new Helm chart values:
+
+   ```console
+   helm upgrade coder coder/coder -n coder --version=<VERSION> --values current-values.yaml
+   ```
+
+1. Upgrade workspaces to use Coder's new Networking v2 functionality.
+
+   If all of your workspaces are local, go to Manage > Providers. Find your
```
bpmct, line 102: A bit confused on what a "local" workspace is. Do you mean using the built-in provider?

Suggest changing to "If you do not have additional workspace providers"

bpmct (changes requested):

bpmct:
> I'm thinking we can dogfood this when upgrading `demo` to `1.21` - thoughts? @mterhar @bpmct

Unsure how/when the demo upgrade happens, but I've done several upgrades to 1.21. I think a lot of the things for this docs are if the person is using a custom ingress. In 1.20 -> 1.21, the built in ingress is updated and then helm instructs users to use serviceNext as the ingress will be removed in future releases.

khorne3:
@bpmct Do you think we should split this guide into two sections where we have one for those using a custom ingress and one for those who aren't?

setup/updating.md
ericpaulsen: This I'm not sure.

bpmct:
> @bpmct Do you think we should split this guide into two sections where we have one for those using a custom ingress and one for those who aren't?

Yeah I think that would make it explain more. For folks using the built-in ingress they shouldn't see any difference but will see a helm message recommending them to upgrade some values (if they didn't pre-upgrade). coderd.serviceNext won't upgrade the ingress, it actually removes it.

setup/updating.md
```diff
@@ -65,8 +65,54 @@ To update Coder, follow these steps:
...
+   Refer to the `1.21`
+   [helm chart](https://github.com/cdr/enterprise-helm/blob/main/values.yaml)
+   when making such changes:
+
+   Rename all instances of `cemanager` to `coderd`.
+
+   Redirect all traffic from your custom ingress to `coderd`.
+
+   `coderd` now serves TLS certificates; to upgrade the NGINX ingress
+   controller, set `coderd.serviceNext` to `true`. (See our
```
bpmct, line 82: I am pretty sure this helm value actually removes the ingress controller and serves Coder directly through a LoadBalancer. cc @kylecarbs to confirm/deny.

setup/updating.md
```diff
@@ -65,8 +65,54 @@ To update Coder, follow these steps:
...
+   >
+   > To manage the Access URL, go to **Manage** > **Admin** >
+   > **Infrastructure**.
+
+1. Migrate the following ports (this may need to be done using your cloud
+   provider):
+
+   Move TCP port `8080` to `80`.
+
+   Move TCP port `8443` to `443`.
```
bpmct, line 98: Where do you migrate the ports? I'm a bit confused by this

bpmct (dismissed):

setup/updating.md
khorne3: If that's the case, we need to update the messaging to people when they upgrade: https://github.com/cdr/enterprise-helm/blob/main/templates/NOTES.txt#L51

setup/updating.md
ericpaulsen: I believe this would need to be done at the Load Balancer level (if there is one). @mterhar please confirm or clarify.

setup/updating.md
khorne3: Yes, custom ingresses

khorne3:
I'm wondering if we should split this into a guide; if you use the Helm chart as we provide it, the upgrade is one additional step to what we suggest; it's only w/ custom ingresses that things become gnarly

ericpaulsen (approved):

ericpaulsen:
We need to push this through. This has remained undocumented for too long, as we have many customers working to upgrade to `1.21x`

PR #530, opened 2021-08-02
cloud DNS docs: fix outdated dns01 solver name

bpmct:
Context: The current docs do not work for someone using the version of cert manager we recommend. I ran into this while testing the cloud DNS docs on a new cluster

The old version we used to recommend works fine.

https://github.com/streamnative/charts/pull/135/files#diff-f446351342bdec56ed4dcd24f2fce3d14fee4c3c50c3bf92e07637b3fac885acL41-L42


ericpaulsen:
Good catch! Was in the process of making this change myself.

ericpaulsen (approved):

khorne3 (approved):

PR #552, opened 2021-08-24
chore: add STUN doc

khorne3:


admin/stun.md
```diff
@@ -0,0 +1,13 @@
+---
+title: "P2P Workspace Connections"
```
jawnsy, line 2:
```suggestion
title: "Direct Workspace Connections"
```

@ashumz @kylecarbs I thought we're using "Direct Connect" as the term now? Although our settings page still does say P2P :thinking::

image

jawnsy (approved):

admin/stun.md
khorne3: I may have made things more confusing by changing this to Peer-to-Peer Connections. But, as long as we have direct connections, p2p connections, and peer-to-peer connections on the page so it shows up in search, it should be fine and I'll update the screenshot as needed.

jawnsy: @khorne3 No worries from my end, just wanted to make sure we use the correct terminology, and I'm not sure what it is anymore. I seem to recall that @ketang had some concerns about the peer-to-peer terminology, and I thought we had renamed the feature to "Direct Connect", but couldn't find the Slack thread :man_shrugging: 

I think Tyler (not on our GitHub yet) owns naming for the Coder product now, so maybe we can check with him?

PR #573, opened 2021-08-31
Remove dashboard container from air-gapped

mterhar:
@khorne3 How do we get this change back into the 1.21 docs?


mterhar:
@khorne3 I found branches for 1.21 and 1.20 so I went ahead and removed dashboard from those branches too. I assume they'll get pulled on the next publishing event.

ericpaulsen (approved):

khorne3:
@mterhar yup, that's correct; the site rebuilds every six hours or so

PR #574, opened 2021-08-31
add programmatic dev URL access

ericpaulsen:
documenting how to programmatically access authenticated dev URLs, per a customer inquiry. tested & validated on my end.

### DIFF
--- a/workspaces/devurls.md
+++ b/workspaces/devurls.md
@@ -71,6 +71,19 @@ If, however, you named the dev URL `reactproject`, then your URL is
 If you access a dev URL that hasn't been created, Coder automatically adds it to
 your dev URL list on the dashboard and sets the access level to **Private**.
 
+## Programmatically accessing dev URLs
+
+If you need programmatic access to authenticated dev URLs (Private,
+Organization, or Authenticated Users), you can do the following:
+
+```console
+# Generate a token
+coder@dev ~ coder tokens create devurl
+<TOKEN>
+# Curl Dev URL
+coder@dev ~ curl --cookie "devurl_session=<TOKEN>" <dev-url>
+```
+
 ## Access via SSH port forwarding
 
 You can also access your server via [SSH port forwarding](ssh.md#forwarding-dev-urls).
--- a/workspaces/devurls.md
+++ b/workspaces/devurls.md
@@ -81,6 +81,7 @@ you can run the following in the terminal:
 # Generate an API token with the Coder CLI
 $ coder tokens create devurl
 <TOKEN>
+
 # Send HTTP requests to the dev URL using the devurl_session cookie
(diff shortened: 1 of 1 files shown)


workspaces/devurls.md
```diff
@@ -71,6 +71,19 @@ If, however, you named the dev URL `reactproject`, then your URL is
...
+## Programmatically accessing dev URLs
+
+If you need programmatic access to authenticated dev URLs (Private,
+Organization, or Authenticated Users), you can run the following commands:
+
+```console
+# Generate a token
+coder@dev ~ coder tokens create devurl
+<TOKEN>
+# Curl Dev URL
```
bpmct, line 83:
```suggestion
# Send HTTP requests to the dev URL with the devurl_session cookie
```

bpmct (approved):
Cool, tested and works great!

Suggested some changes to clarify:
- coder-cli is only needed once to generate an API token and doesn't have to be on every device making the request.
- CURL is one way to send an http request, the devurl_session cookie is the important part

ericpaulsen:
@bpmct great feedback - I'll add these in.

khorne3 (approved):

workspaces/devurls.md
```diff
@@ -71,6 +71,22 @@ If, however, you named the dev URL `reactproject`, then your URL is
...
+access levels set to **private**, **organization**, or **authorized users**),
+you can run the following in the terminal:
+
+```console
+# Generate an API token with the Coder CLI
+$ coder tokens create devurl
+<TOKEN>
+
+# Send HTTP requests to the dev URL using the devurl_session cookie
+$ curl --cookie "devurl_session=<TOKEN>" <dev-url>
```
jawnsy, line 86: We might want to use a header here over a cookie, as I think cookies have CSRF protections applied. @Emyrk what do you think?

Emyrk: Devurls do not have CSRF protection, as then it would likely break any app being served over the devurl.

The CSRF is after the devurl handler: https://github.com/cdr/m/blob/8ad3c9675faeeb6ad82f1607e08d905b6a3abe0a/product/coder/service/coderd/routes.go#L52-L62

So the cookie is fine, although a header still might be cleaner

jawnsy: @Emyrk Nice, good to know, thank you for the correction. Do we support devurl sessions as a header? If so, what's the header name?

sreya: We don't support headers currently because the browser was the only workflow being considered. It'd be pretty easy to add header support though if we need to.

jawnsy: @sreya Cool. Well if a cookie is fine from a security perspective, then I guess there's no reason to do a header :man_shrugging:

PR #579, opened 2021-09-03
Revert "clarify CVM auth capabilities (#578)"

sreya:
This reverts commit 71be8a627d975ad64cbfece44c94893253cdd1b2.

- There was some confusion on whether CVMs can pull from private
  registries and this commit was attempted to clarify the documentation.
  The truth is CVMs can pull private image fine they just cannot use
  the node credentials.

### DIFF
--- a/admin/workspace-management/cvms.md
+++ b/admin/workspace-management/cvms.md
@@ -141,8 +141,8 @@ container is what provides
 
 ## Images hosted in private registries
 
-Please note that _non-cached_ CVM workspaces cannot be created using images hosted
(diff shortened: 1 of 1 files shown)


ericpaulsen:
If this is the case, then:

```console
Please note that CVM-enabled workspaces cannot be created using images hosted in
 a private registry unless you permit unauthenticated access to the images.
 ```
 
 is incorrect.

sreya:
I don't think reverting this is fair as it seems previous documentation states that we cannot pull private images which is just wrong. I'll open a separate PR that fixes things.

PR #588, opened 2021-09-10
refresh custom images guides in docs

bpmct:
This PR proposes a few things:

1. Remove `guides/customization/custom-workspaces.md`
    - we have a custom images page under "Images > Writing"
    - this guide builds an image from scratch. we usually recommend using our base image.
    - we also have a node.js projects example, which is very similar to this guide
    - the "Images > Structure" includes these dependencies & requirements as the base images
2. change ordering of Images children
    - "Image > Structure" only applies to people building custom images, so I think it should probably be after
3. mention Coder workspaces' persistent volume limitation when creating images [sc-11937]
    - while persistent home volume is great for developer workflow, it has been confusing to some users when building images, as any software or files installed in `/home/user` is not present when the workspace is started. This line should explain why this is happening and how to work around it
4. add concrete examples to the top of "Custom image creation" the page for folks who may not have a lot of experience with Docker
    - the Node.js guide includes the docker commands and which files are needed.

Tried to give context so that these don't seem like a bunch of "random" changes. Would appreciate feedback on how this could be done better, or differently :)


khorne3 (approved):

jsjoeio (approved):
Woohoo! Nice work, Ben 🚀

PR #601, opened 2021-09-13
check for 404s and 500s on previews

bpmct:


.github/workflows/preview.yaml
```diff
@@ -61,6 +61,16 @@ jobs:
           VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
           CODER_DOCS_MAIN_BRANCH: ${{ github.event.pull_request.head.sha }}
 
+      # This ensures the docs site is built properly
+      # as it will sometimes throw a 404 or 500
+      # error if Markdown/parsing errors are present
+      - name: Install node_modules
+        run: yarn install
+      - name: Check docs
+        run: yarn ts-node ./product/coder.com/site/scripts/checkDocs.ts
```
jawnsy, line 70: IMHO it's a little clearer to use `yarn run ts-node` because people less familiar with Node might not be familiar with this behavior.

I'd also suggest putting this at the end of the build (after comment credentials) so that people can see the link to the preview. Otherwise, people won't be able to easily inspect the deployment (needed if you want to open the Vercel console and watch the function logs while you bring up the docs site)

BrunoQuaresma (approved):
LGTM.

PR #603, opened 2021-09-14
fix issue w/ preview config

khorne3:
forgot some quotes

### DIFF
--- a/.github/workflows/preview.yaml
+++ b/.github/workflows/preview.yaml
@@ -4,7 +4,7 @@ on:
   pull_request:
     branches:
       - main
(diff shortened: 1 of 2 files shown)


BrunoQuaresma (approved):
LGTM

.github/workflows/preview.yaml
```diff
@@ -4,7 +4,7 @@ on:
   pull_request:
     branches:
       - main
-      - release-*
+      - 'release-*'
```
BrunoQuaresma, line 7: Would be interesting to use the double quotes to keep the pattern

PR #604, opened 2021-09-14
chore: add Doctor to installation instructions

khorne3:


setup/installation.md
```diff
@@ -17,6 +17,14 @@ Install the following dependencies if you haven't already:
 [PostgreSQL](https://www.postgresql.org/docs/12/admin.html) instance to store
 data, including workspace information and session tokens.
 
+## Doctor
+
+Before proceeding, we strongly recommend running Doctor, which is Coder-supplied
```
johnstcn, line 22:
```suggestion
Before proceeding, we strongly recommend running [Doctor](https://github.com/cdr/coder-doctor), which is Coder-supplied
```

khorne3: @johnstcn this repo is currently private; will we be making it public?

johnstcn: As far as I know, that's the plan!

bpmct: I think we should probably include instructions on how to download from GitHub, unzip, and run Doctor. We can either do this on the doctor repo, or in docs.

Rationale is that our docs include all helm commands, etc. Install steps for Doctor will be helpful

PR #623, opened 2021-09-20
chore: add mention of service account name selector

khorne3:


admin/workspace-providers/management.md
```diff
@@ -47,11 +47,15 @@ At this point, you can:
...
   > before you remove them from a workspace provider's allowlist.
 
-- Change the features of the workspace provider. You can enable **end-to-end
-  encryption** for this provider, enable **external SSH connections** to the
-  provider's workspaces via the Coder CLI, or specify a **Kubernetes storage
-  class** to use when Coder provisions workspaces (this is useful for improving
-  disk performance).
+- Change the features of the workspace provider. You can:
+
+  - Enable **end-to-end encryption** for this provider
```
ericpaulsen, line 52: I'm not seeing where to enable this feature when editing a provider. Am I missing something?

khorne3: I'm looking in master -- is that what you're using, too?

ericpaulsen: I was looking on both demo & stable, which are on 1.22 & 1.23, respectively. Master is on an older version (v1.21.0-rc.1+566-gf0de9bff4-20210917), so for this 1.22 version of docs, we might need to remove this piece.

khorne3: Weird -- I just deployed a fresh copy of `v1.23.0-rc.1` and I see it in my test deployment.

I'd intended these docs to go into v1.23, but it looks like the feature shipped earlier?

PR #625, opened 2021-09-21
chore: remove upgrade instructions relevant to 1.20 --> 1.21

khorne3:
At some point, I think we should remove the instructions required for the Helm chart changes going from 1.20 --> 1.21. If we think we should keep them for the moment, I can close this PR. Thoughts?


ericpaulsen:
I agree that at some point we'll need to deprecate these instructions. But for now, I think we should keep them. Maybe once we hit 1.24 or 1.25 we can consider removing.

PR #636, opened 2021-09-28
 updates to VNC guide for v1.23

bpmct:
Honestly, forgot we had this as well as the VNC sample image


bpmct:
fix for broken preview in #638

PR #638, opened 2021-09-28
fix previews: use node 14

bpmct:
Context: Previews are failing now that coder.com uses Node v14.

If you are not using Coder to edit docs, you'll need to update to Node v14 for pre and post-commit hooks to work.


jawnsy:
This appears to be a duplicate of https://github.com/cdr/docs/pull/635

bpmct:
🤦🏼‍♂️

PR #647, opened 2021-10-05
hotfix(rstudio): add correct frame options

Emyrk:
# What this does

This fixes the `X-FRAME-OPTIONS` to be set to `same`. This is required to fix rstudio in 1.23. Any previous versions, this has no effect.

### DIFF
--- a/workspaces/editors.md
+++ b/workspaces/editors.md
@@ -230,6 +230,7 @@ RStudio:
    RUN chown -R coder:coder /var/lib/rstudio-server
    RUN echo "server-pid-file=/tmp/rstudio-server.pid" >> /etc/rstudio/rserver.conf
    RUN echo "server-data-dir=/tmp/rstudio" >> /etc/rstudio/rserver.conf
+   RUN echo "www-frame-origin=same" >> /etc/rstudio/rserver.conf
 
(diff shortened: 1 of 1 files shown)


vapurrmaid:
[sc-17709]

workspaces/editors.md
```diff
@@ -230,6 +230,7 @@ RStudio:
    RUN chown -R coder:coder /var/lib/rstudio-server
    RUN echo "server-pid-file=/tmp/rstudio-server.pid" >> /etc/rstudio/rserver.conf
    RUN echo "server-data-dir=/tmp/rstudio" >> /etc/rstudio/rserver.conf
+   RUN echo "www-frame-origin=same" >> /etc/rstudio/rserver.conf
```
vapurrmaid, line 233: @khorne3 is there a way to add this to 1.23 as well?

vapurrmaid (approved):

workspaces/editors.md
Emyrk: Yes, we can add this to any docs. It's valid even for older versions that do not need it

khorne3:
@vapurrmaid 1.23 is generated using `main` so we're good to go!

khorne3 (approved):

PR #657, opened 2021-10-11
chore: format index.md

BrunoQuaresma:
Format index.md to be compatible with the Babel parser which is required for some newest mdx parsers that we want to use in the future.


BrunoQuaresma:
Good catch! This is unexpected... I will see how I can fix that. Thank you.

PR #659, opened 2021-10-11
Fixed the position of <children>

mukundshah:
image
The section surrounded in red was supposed to be above the `Incompatible Kubernetes distribution`
I tried to fix this.


vapurrmaid:
Thanks for finding and fixing this -> @all-contributors please add @mukundshah for bug

PR #673, opened 2021-10-19
chore: add usage note for Azure users

khorne3:


guides/admin/timescale-migration.md
```diff
@@ -10,7 +10,14 @@ At the end of your evaluation period, you may need to migrate the data from the
...
 This article will walk you through the process of doing so.
 
-> You must be a cluster admin for your Kubernetes cluster.
+## Requirements
+
+- You must be a cluster admin for your Kubernetes cluster.
+
+- **Azure database users**: if you're using Azure Database for PostgreSQL, note
+  that Coder only works with Single Server; the Flexible Server (Preview) and
+  Hyperscale (Citus) options do _not_ support the TimescaleDB extension
```
jawnsy, line 19: Wait, I didn't realize we were using Timescale features still, my understanding is that we're using stock PostgreSQL completely now. How did we run into this issue?

cc @coadler

jawnsy: @khorne3 @mterhar Wait, I didn't realize we were using Timescale features still, my understanding is that we're using stock PostgreSQL completely now. How did we run into this issue?

cc @coadler @ericpaulsen

PR #675, opened 2021-10-19
chore: update node-group info for CVMs on AWS

khorne3:


khorne3:
@ericpaulsen this is the CVMs doc, not the AWS K8s doc

ericpaulsen (approved):

PR #676, opened 2021-10-19
feat: add guide for running Tailscale in Coder

jawnsy:
Preview link: https://codercom-4s8229gi9-codercom.vercel.app/docs/coder/latest/guides/customization/tailscale


guides/customization/tailscale.md
```diff
@@ -0,0 +1,278 @@
...
+
+[contact our support team]: ../../feedback.md
+[container-based virtual machine]: ../../workspaces/cvms.md
+
+## Step 1: Create the Dockerfile
+
+In Coder, developer workspaces are defined by a Dockerfile that contains the
+apps, tools, and dependencies that you need to work on the project.
+
+> See Docker’s [guide to writing Dockerfiles] for more information.
```
bpmct, line 51: I suggest also linking to [our guide on creating Dockerfiles](https://coder.com/docs/coder/latest/images/writing#creating-a-custom-image), as Coder images require a differ
```suggestion
> See our [custom image docs](../../images/writing) and
> Docker’s [guide to writing Dockerfiles] for more information.
```

guides/customization/tailscale.md
```diff
@@ -0,0 +1,278 @@
...
+
+Create the folder hierarchy using:
+
+```console
+$ mkdir --parents --verbose \
+    files/etc/apt/preferences.d \
+    files/etc/apt/sources.list.d \
+    files/etc/systemd/system/tailscaled.service.d \
+    files/usr/share/keyrings
+mkdir: created directory 'files'
```
bpmct, line 92: Suggest adding a newline  to separate the command and the output
```suggestion
    files/usr/share/keyrings

mkdir: created directory 'files'
```

guides/customization/tailscale.md
```diff
@@ -0,0 +1,278 @@
...
+        └── share
+            └── keyrings
+                └── tailscale.gpg
+```
+
+While it is possible to configure everything directly in a single Dockerfile, we
+recommend using a folder hierarchy, since it makes it easier to create a
+reproducible image and examine the change history for individual files.
+
+Create the folder hierarchy using:
```
bpmct, line 84: Wondering if there should also be a command to create the files after the folders, I was slightly confused that I had to create the files, but it wasn't a huge deal.

```sh
touch files/etc/apt/preferences.d/tailscale \
    files/etc/apt/sources.list.d/tailscale.list \
    files/etc/systemd/system/tailscaled.service.d/tailscale.conf
```

khorne3 (approved):

guides/customization/tailscale.md
```diff
@@ -3,45 +3,51 @@ title: Tailscale
...
+
+As part of this tutorial, you'll create an image with the following that you'll
+use to create new Coder workspaces. The container image will include:
+
+- The Tailscale daemon (`tailscaled`)
+- A transparent proxy tool (`proxychains4`)
 - Environment variables to control proxy behavior for outbound connections
-- A [SOCKS5 proxy] listening for connections on `localhost:1080`
-- A [HTTP proxy] listening for connections on `localhost:3128`
+- A [SOCKS5 proxy] to listen for connections on `localhost:1080`
```
jawnsy, line 22: I think I phrased it as "listening for" because the tailscale daemon is listening for connections, this phrasing seems to imply that there will be a separate proxy listening for connections... do you have suggestions for a clearer/better way to convey what I intended?

guides/customization/tailscale.md
```diff
@@ -78,48 +84,37 @@ the target files, in a subdirectory called `files`:
...
-recommend using a folder hierarchy, since it makes it easier to create a
+recommend using a folder hierarchy, since this makes it easier to create a
 reproducible image and examine the change history for individual files.
 
-Create the folder hierarchy using:
+Create the folder hierarchy:
 
 ```console
-$ mkdir --parents --verbose \
+mkdir --parents --verbose \
```
jawnsy, line 93: if we remove the output, let's remove --verbose from the example too (console and shell-session are meant to show the terminal prompt, input command, and command results, so that users can confirm that the command ran as we expect)

I can't find a great reference right now, but did find this https://stackoverflow.com/a/49004070

guides/customization/tailscale.md
```diff
@@ -78,48 +84,37 @@ the target files, in a subdirectory called `files`:
...
 
 Add the Tailscale package repository to `tailscale.list` in the local path
-`files/etc/apt/sources.list.d`, which will appear in `/etc/apt/sources.list.d`
-in the resulting image, with the following contents:
+`files/etc/apt/sources.list.d`. This will appear in `/etc/apt/sources.list.d` in
+the resulting image, with the following contents:
 
 ```text
-# Tailscale packages for ubuntu focal
+# Tailscale packages for Ubuntu Focal Fossa
```
jawnsy, line 107: the comment came from the Tailscale download instructions, we could remove the comment entirely

https://tailscale.com/download

guides/customization/tailscale.md
```diff
@@ -133,29 +128,30 @@ Pin: origin pkgs.tailscale.com
...
+Coder). We will need to modify the service settings to:
 
 - Store the state file in the persistent home volume (`/home/coder`)
 - Enable userspace networking
 - Enable the SOCKS5 proxy (optional)
 - Enable the HTTP proxy (optional)
 
 If you do not require outbound connections from the workspace to other services
-running in the tailnet, you may skip configuring the proxies.
+running in the tailnet, you may skip the final two steps configuring the
```
jawnsy, line 153: we'll have to be careful if we add steps to also modify this :thinking: is there a good way to phrase this that doesn't refer to the position?

guides/customization/tailscale.md
```diff
@@ -200,11 +196,12 @@ ENV http_proxy=http://localhost:3128
...
 ## Step 2: Authenticate to Tailscale
 
-Start a workspace using the container image. Initially, `tailscaled` should be
-running, but will indicate that it requires authentication:
+[Create a workspace] using the container image. Initially, `tailscaled` should
+be running, but it will indicate that it requires authentication:
 
 ```console
-$ systemctl status tailscaled
+systemctl status tailscaled
```
jawnsy, line 203: I think the `$` prefix is important to distinguish that: (1) this command is not run as root; and (2) to distinguish the prompt from the example output

but if our convention is to skip the example output or put the output in a separate fenced code block, we can do that

guides/customization/tailscale.md
jawnsy: Dunno if it helps or makes things worse, [tailscale has a diagram](https://tailscale.com/kb/1112/userspace-networking/):

image

Does this makes sense to you? Networking is always awful to explain, so I probably did a bad job of it, sorry!

jawnsy: Ooo, I like that phrasing. Maybe "An [HTTP proxy] included in tailscaled to facilitate connections to the tailnet, listening on ... "?

PR #698, opened 2021-11-05
chore: add explanation to user roles

khorne3:
I played with a bunch of different charts, including [this one](https://docs.google.com/spreadsheets/d/1bQG5GTZML3lCQPpKMAEy2pBIeY40_eJkK6h9pWPuLF0/edit?usp=sharing), but in the end, I didn't think the added complexity would be helpful to anyone reading our user perms doc. So, I just added a brief summary of how the permissions are additive (mostly).

Will also add the automated tables when we have that ready.

### DIFF
--- a/admin/access-control/user-roles.md
+++ b/admin/access-control/user-roles.md
@@ -39,6 +39,20 @@ There are four roles available:
     </tbody>
 </table>
 
+### Additive permissions
+
+The following tables detail what permissions Coder grants to each of the four
+roles, but a summary of the roles are:
+
+- All users are (or have the permissions of) a **member**
+- An auditor has the permissions of a member, plus the ability to work with
+  audit logs
+- A site manager has the permissions of a member or an auditor, please
+  additional administrative rights
+- A site admin has the permissions of a member, auditor, and site manager, as
+  well as additional admin rights (e.g., creating site managers, access to API
+  keys).
+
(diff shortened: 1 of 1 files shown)


admin/access-control/user-roles.md
```diff
@@ -39,6 +39,20 @@ There are four roles available:
...
 
+### Additive permissions
+
+The following tables detail what permissions Coder grants to each of the four
+roles, but a summary of the roles are:
+
+- All users are (or have the permissions of) a **member**
+- An auditor has the permissions of a member, plus the ability to work with
+  audit logs
+- A site manager has the permissions of a member or an auditor, please
```
jawnsy, line 50:
```suggestion
- A site manager has the permissions of a member or an auditor, plus
```

admin/access-control/user-roles.md
```diff
@@ -39,6 +39,20 @@ There are four roles available:
...
+  additional administrative rights
+- A site admin has the permissions of a member, auditor, and site manager, as
+  well as additional admin rights (e.g., creating site managers, access to API
+  keys).
```
jawnsy, line 54:
```suggestion
  keys)
```

removed the full stop for consistency with other bullets

jawnsy (approved):
LGTM!

PR #722, opened 2021-11-30
removed carriage returns and changed title to use industry term Gateway

mtm20176:
For some reason, there were carriage returns on steps 6, 7. Also changed the section title to state Gateway since that is the term prospects know who use JetBrains.


khorne3 (approved):

bpmct (commented):
LGTM

PR #726, opened 2021-12-01
chore: update GPU doc

khorne3:


admin/workspace-management/gpu-acceleration.md
```diff
@@ -3,22 +3,22 @@ title: "GPU acceleration"
...
-workspaces.
+Graphical processing units (GPUs) are useful with compute-intensive workloads,
+such as those involved with data science/machine learning projects.
 
-By default, users cannot create workspaces with GPUs until a site manager
-configures and enables this feature.
+You can allocate GPUs to workspaces once a site manager configures and enables
+this feature. Enabling GPU acceleration requires that you've configured your
+Kubernetes cluster appropriately and
+[scheduled GPUs](https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/).
```
ericpaulsen, line 12: I believe this should be `schedules GPUs`.

admin/workspace-management/gpu-acceleration.md
```diff
@@ -3,22 +3,22 @@ title: "GPU acceleration"
...
 
-### Step 1: Configure your Kubernetes cluster
+- [Amazon Elastic Container Service](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-gpu.html)
```
ericpaulsen, line 14: Maybe we should add a heading before just listing the cloud providers? The text doesn't seem to provide much of a description as to what is being linked to. Just a thought.

admin/workspace-management/gpu-acceleration.md
khorne3: Think the past tense is correct, but removed the extra words that made this confusing

ericpaulsen (approved):

PR #730, opened 2021-12-03
chore: add K8s version support chart

khorne3:


setup/kubernetes/index.md
```diff
@@ -25,10 +25,16 @@ stable version. We follow this policy to ensure that Coder stops using
 deprecated features before they are removed from new versions of Kubernetes.
 
 <!-- markdownlint-disable -->
+
+| **Coder**                         | v1.25                        | v1.24                        | v1.23                        | v1.22                        |
+| --------------------------------- | ---------------------------- | ---------------------------- | ---------------------------- | ---------------------------- |
+| **Supported Kubernetes versions** | `1.22.x`, `1.21.x`, `1.20.x` | `1.21.x`, `1.20.x`, `1.19.x` | `1.21.x`, `1.20.x`, `1.19.x` | `1.20.x`, `1.19.x`, `1.18.x` |
```
jawnsy, line 31: I think `1.22`, `1.21`, `1.20` (i.e. drop the `.x` suffix) would be enough to convey that we support all patch versions of these minor versions

Alternatively, maybe we could have a tabular form with Kubernetes versions in the rows on the left? Something like this:

|            | Kubernetes 1.22 | Kubernetes 1.21 | Kubernetes 1.20 | Kubernetes 1.19 | Kubernetes 1.18 |
|------------|-----------------|-----------------|-----------------|-----------------|-----------------|
| Coder 1.25 | ✅              | ✅              | ✅              |                 |                 |
| Coder 1.24 |                 | ✅              | ✅              | ✅              |                 |
| Coder 1.23 |                 | ✅              | ✅              | ✅              |                 |
| Coder 1.22 |                 |                 | ✅              | ✅              | ✅              |

Something that this also gets into is our support policy for Coder itself, i.e. which versions we will backport fixes to, and I'm not sure where we landed on that. This table as-is implies that we are actively providing security fixes for 1.22, 1.23, 1.24, 1.25, and I'm not sure whether that is true.

I put Kubernetes in the columns and Coder in the rows because Kubernetes releases at a slower cadence (previously every 3 months, [and now every 4 months](https://kubernetes.io/blog/2021/07/20/new-kubernetes-release-cadence/)) than Coder (which releases monthly), so I figured that there will generally be more rows than columns.

khorne3: > which versions we will backport fixes to, and I'm not sure where we landed on that

I think we've been backporting to as early as 1.19 in the recent weeks

jawnsy (approved):

PR #731, opened 2021-12-03
Bump advertised C4D to 1.25.1

Emyrk:
# What this does

Only coder for docker has the 1.25.1. Normal customers do not update. This is a C4D patch only


deansheather (approved):

jawnsy:
[sc-19483]

jawnsy:
Looks like this was superseded by #732

PR #748, opened 2021-12-15
fix: Coder image workflow wasn't working

bpmct:
`tj-actions/changed-files` was not being used, so I removed it.

```
Error : .github#L1
tj-actions/changed-files@v2.1 is not allowed to be used in cdr/docs. Actions in this workflow must be: within a repository owned by cdr, created by GitHub, verified in the GitHub Marketplace or match the following: marocchino/sticky-pull-request-comment@*, styfle/cancel-workflow-action@*, w3f/action-find-old-files@*, 8398a7/action-slack@*.
```

Uses the proper image name as well.

### DIFF
--- a/.github/workflows/build_coder_image.yaml
+++ b/.github/workflows/build_coder_image.yaml
@@ -10,9 +10,6 @@ jobs:
   build-image:
     runs-on: ubuntu-latest
     steps:
-      - name: Get changed files
-        id: changed-files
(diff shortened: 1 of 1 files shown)


khorne3 (approved):

jawnsy:
@bpmct Thanks for noticing and fixing that! BTW, it's in an allowlist setting in the repo, just helps make things more secure because it prevents people from adding arbitrary actions to our builds

PR #762, opened 2021-12-21
chore: fix title casing

khorne3:


mtm20176 (changes requested):
You need to specific 'Open' in front of VSX's

mtm20176 (approved):

PR #764, opened 2021-12-21
chore: update air gapped deploy command & values

ericpaulsen:
making this change to include the [example helm values](https://github.com/coder/enterprise-helm/blob/main/examples/offline/offline.values.yaml) created by @jawnsy. [slack thread here](
https://codercom.slack.com/archives/C0102H1BBKK/p1640115155082700).

### DIFF
--- a/setup/air-gapped/index.md
+++ b/setup/air-gapped/index.md
@@ -117,25 +117,25 @@ platform images are hosted in Coder's Docker Hub repo.
    docker push my-registry.com/coderenvs/coder-service:<version>
    ```
 
-1. Create an `offline.values.yaml` file which includes the image paths for each
-   the Coder containers and proxy configuration (if necessary). Below is an
-   example:
+1. If necessary, create an `offline.values.yaml` file that includes the image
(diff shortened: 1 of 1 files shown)


setup/air-gapped/index.md
```diff
@@ -117,15 +117,32 @@ platform images are hosted in Coder's Docker Hub repo.
...
+   # Coder will use this proxy for all outbound HTTP/HTTPS connections
+   # such as when checking for updated images in the image registry.
+   # However, note that images are pulled from the Kubernetes container runtime,
+   # and may require a different setting.
+      proxy:
+         http: http://proxy.internal:8888
+         exempt: cluster.local
+   postgres:
+      default:
+         image: my-registry.com/coderenvs/timescale:<version>
```
jawnsy, line 136: all the images should use a similar path as that's the most common case, suggest changing the envbox one to match

jawnsy (approved):
Thanks for this!

khorne3 (approved):

PR #767, opened 2021-12-22
chore: update EKS troubleshooting steps

ericpaulsen:
adding a second option to troubleshoot LB health check failures in EKS. 

since this is caused by `coderd` being scheduled on the additional node group we ask users to deploy, the second option has users schedule `coderd` on the EKS `standard-workers` node group via the `services.nodeSelector` helm value.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "@types/semver-compare": "^1.0.1",
     "all-contributors-cli": "^6.20.0",
     "husky": "^7.0.1",
-    "lint-staged": "^12.0.2",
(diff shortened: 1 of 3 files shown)


jawnsy (approved):

khorne3 (approved):

ericpaulsen:
is there a way to get around this lint check? the markdown numbering did not add `2.` for the second option, so I hard-coded it.

khorne3:
should be all good now @ericpaulsen

PR #791, opened 2022-01-12
fix: remove dev url var quotations

ericpaulsen:
removing the `DEVURL_HOST` quotations in the Docker Compose file. a prospect reported this threw an error where unnecessary characters were added to newly created dev URLs.


johnstcn (approved):
Nice catch @jawnsy!

PR #809, opened 2022-01-21
add ebpf-required config

ericpaulsen:
discovered a particular kernel configuration that is needed to successfully run the `exectrace` sidecar container:

- `CONFIG_DEBUG_INFO_BTF=y`

@deansheather let me know if we should include the command(s) to run to check for this on the host system.

### DIFF
--- a/admin/workspace-management/process-logging.md
+++ b/admin/workspace-management/process-logging.md
@@ -39,7 +39,7 @@ machine in question:
 cat /proc/config.gz | gunzip | grep CONFIG_DEBUG_INFO_BTF
 ```
 
-```
+```console
 cat "/boot/config-$(uname -r)" | grep CONFIG_DEBUG_INFO_BTF
 ```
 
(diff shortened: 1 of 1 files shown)


deansheather (approved):
We should probably put the commands I posted in Slack so people can check this

admin/workspace-management/process-logging.md
```diff
@@ -30,7 +30,18 @@ The core of this feature is also open source and can be found in the
 ## Requirements
 
 Use of the workspace process logging functionality requires a host Linux
-kernel >= 5.8.
+kernel >= 5.8 with the kernel config `CONFIG_DEBUG_INFO_BTF=y` enabled.
+
+To validate this config is enabled, run either of the following commands on the
+machine in question:
```
deansheather, line 36:
```suggestion
nodes directly (not from within a workspace):
```

deansheather (approved):

PR #820, opened 2022-01-28
chore: rename Create a Workspace doc

khorne3:


setup/docker.md
```diff
@@ -56,9 +56,9 @@ Coder requires x86-64 and does not support ARM-based processors at this time.
 1. Launch a web browser and navigate to the URL provided by Coder (e.g.,
    `http://localhost:7080`). Log in using the credentials Coder provided.
 
-1. At this point, you can [create a workspace](../workspaces/getting-started.md)
-   using one of the **Packaged** images by clicking on **New workspace** in the
-   center of the UI.
+1. At this point, you can [create a workspace](../workspaces/create.md) using
```
jawnsy, line 59: This will only be available for installations that are not air-gapped. I'd suggest we add something to "Known issues" (or maybe call it "Limitations") at the bottom of the page indicating that C4D does not support air-gapped/offline installs

jawnsy (approved):

deansheather (approved):

PR #824, opened 2022-02-01
Update OIDC Docs

hewittwill:
It's no longer a pre-requisite to be setup for the Identity Platform, and you need to prefix `accounts.google.com` with `https://`

### DIFF
--- a/guides/admin/oidc-google.md
+++ b/guides/admin/oidc-google.md
@@ -16,12 +16,6 @@ feature requires you to provide three pieces of information from Google:
 This guide will show you how to set up an app on Google and obtain the
 information you need to provide to Coder.
 
-## Prerequisites
-
-Before proceeding, please ensure that you've
-[enabled and configured the Identity Platform](https://cloud.google.com/identity-platform/docs/web/oidc)
-for your Google Cloud account.
-
 ## Step 1: Create the OAuth consent screen
 
 1. Navigate to your [GCP console](https://console.cloud.google.com).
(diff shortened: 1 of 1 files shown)


guides/admin/oidc-google.md
```diff
@@ -16,12 +16,6 @@ feature requires you to provide three pieces of information from Google:
 This guide will show you how to set up an app on Google and obtain the
 information you need to provide to Coder.
 
-## Prerequisites
-
-Before proceeding, please ensure that you've
```
jawnsy, line 21: > It's no longer a pre-requisite to be setup for the Identity Platform, and you need to prefix accounts.google.com with https://

@hewittwill Do you have a reference for these? Our account has Identity Platform enabled, so I don't have an easy way to verify this

The linked page refers to configuring login using Identity Platform, so it might be that Google supports multiple ways to configure it as an IdP for Coder (e.g. using Google Workspace, using Cloud Identity, or Identity Platform) depending on the usecase.

Please correct me if I'm wrong - my current understanding is that these are all separate identity products:

1. Google Workspace logins are for employees who have a GSuite/Workspace account
2. Cloud Identity is used by people who do not use Google Workspace (or for managed accounts that exist outside of the Google Workspace - e.g. contractors that have a login but do not have access to other Workspace services), or companies that use another IdP (e.g. Azure Active Directory) as their primary SSO system
3. Identity Platform is used by people for customer accounts (as opposed to employee/contractor accounts), since it's a Customer Identity and Access Management system, similar to Okta

PR #828, opened 2022-02-03
chore: add missing node label step

jawnsy:
With the example deployment manifest for smarter-device-manager,
users need to label compatible nodes in order for the DaemonSet
to run.

### DIFF
--- a/guides/deployments/podman.md
+++ b/guides/deployments/podman.md
@@ -142,8 +142,27 @@ Linux kernel doesn't support CVMs.
    templates. To do so, go to **Manage** > **Admin** > **Templates**, and set
    the **Enable workspace templates** to **On**. Click **Save**.
 
-1. Create a YAML file with the following contents (the instructions ask the
-   cluster to request the FUSE device for each workspace):
+1. Create a
+   [workspace configuration file](../../workspaces/workspace-templates/templates.md)
+   that includes instructions for resource requests and resource limits (the
+   instructions ask the cluster to request the FUSE device for each workspace):
+
+   ```yaml
+   version: "0.2"
+   workspace:
+     specs:
+       kubernetes:
+         resource-requests:
+           policy: write
+           value:
+             smarter-devices/fuse: "1"
+         resource-limits:
+           policy: write
+           value:
+             smarter-devices/fuse: "1"
+   ```
+
+   A complete workspace template might look something like
 
    ```yaml
    version: "0.2"
--- a/guides/deployments/podman.md
+++ b/guides/deployments/podman.md
@@ -14,6 +14,18 @@ Linux kernel doesn't support CVMs.
 > Please be aware that there are
 > [limitations related to running Podman in rootless mode](https://github.com/containers/podman/blob/main/rootless.md#shortcomings-of-rootless-podman).
 
+> **Note**: Running Podman in rootless mode requires a FUSE device to implement
+> the overlay filesystem (fuse-overlayfs) in unprivileged mode. The following
+> directions work by mounting the FUSE device from the host into workspace
+> containers, which conflicts with the isolation provided by SELinux and
+> AppArmor.
+>
+> For systems running AppArmor (typically Debian- and Ubuntu-derived systems),
+> please disable AppArmor before proceeding.
+>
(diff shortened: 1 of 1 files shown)


guides/deployments/podman.md
```diff
@@ -14,6 +14,18 @@ Linux kernel doesn't support CVMs.
 > Please be aware that there are
 > [limitations related to running Podman in rootless mode](https://github.com/containers/podman/blob/main/rootless.md#shortcomings-of-rootless-podman).
 
+> **Note**: Running Podman in rootless mode requires a FUSE device to implement
```
jawnsy, line 17: the linter doesn't like having two adjacent blockquote sections. We could make this top-level text or combine it with the above (since these are really limitations of the approach). I just wanted to describe in greater depth why this method is incompatible with SELinux/AppArmor

khorne3 (approved):

guides/deployments/podman.md
```diff
@@ -137,69 +150,85 @@ Linux kernel doesn't support CVMs.
...
-           start:
-               policy: write
-       dev-urls:
+     configure:
+       start:
+         policy: write
+     dev-urls:
+       policy: write
+     specs:
+       aws-ec2-docker:
```
jawnsy, line 159: whoops, I dunno how the stuff with ec2 and docker slipped in, but I don't think we need those?

khorne3: weird, I'll fix. sorry!

khorne3: ok, I'm not in the right headspace to fix this right now. will return to this in the AM.

jawnsy: sounds good, thanks Katie!

guides/deployments/podman.md
```diff
@@ -137,69 +148,85 @@ Linux kernel doesn't support CVMs.
...
+         privileged:
+           policy: read
+         resource-requests:
+           policy: write
+           value:
+             smarter-devices/fuse: "1"
+         resource-limits:
+           policy: write
+           value:
+             smarter-devices/fuse: "1"
```
jawnsy, line 199: These lines are really the only relevant ones, so if we had some way to highlight these (to indicate to users that this is what is needed to mount the FUSE device needed to run Podman), then that would clarify what this template is showing

guides/deployments/podman.md
```diff
@@ -127,79 +128,124 @@ Linux kernel doesn't support CVMs.
...
+   ```
+
+   A complete workspace template might look something like
```
khorne3, line 165: wdyt @jawnsy ?

guides/deployments/podman.md
```diff
@@ -127,79 +128,124 @@ Linux kernel doesn't support CVMs.
    kubectl apply -f ./smarter-device-manager.yaml
    ```
 
+   The example `DaemonSet` includes a `nodeSelector` that constrains the device
+   plugin to nodes with the `smarter-device-manager` label set to `enabled`.
+   Label the nodes that will include the FUSE device by using the following
+   command, or remove the `nodeSelector` from the manifest:
+
+   ```console
+   kubectl get nodes
```
jawnsy, line 137: this was my fault and it's pretty minor but this code block gets a Copy button, but the resulting snippet isn't useful on its own since the node name needs to be replaced.

Looks like we can replace this with the following, to label all nodes accordingly: `kubectl label nodes --all smarter-device-plugin=enabled`

guides/deployments/podman.md
jawnsy: Awseome idea! :brain: 

Thanks Katie. Looks great to me

guides/deployments/podman.md
jawnsy: @khorne3 my bad, I gave you bad directions in the above comment, the code I had was right but the comment was wrong

```
         nodeSelector:
           smarter-device-manager: enabled
```

This means the label should be `smarter-device-manager=enabled` instead of `smarter-device-plugin` :man_facepalming:

jawnsy:
Took another quick look through, LGTM! Merging. Thanks for your help editing @khorne3 !

PR #830, opened 2022-02-04
feat: document proxy settings

jawnsy:
Document use of the new Helm chart settings for forward and
reverse proxies.

--

I'm not attached to any of this, please feel free to edit mercilessly! :)


manifest.json
```diff
@@ -440,6 +440,9 @@
             {
               "path": "./guides/deployments/podman.md"
             },
+            {
+              "path": "./guides/deployments/proxy.md"
```
jawnsy, line 444: I'm open to suggestions for where to put this guide or what to name it - please feel free to rename it and push to this branch.

Also, it seems that this list is not alphabetically ordered (`postgres` would come after `podman` in that case), but I assume that's intentional?

guides/deployments/proxy.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+If you specify a proxy for outbound HTTP connections and do not specify a proxy
+for outgoing HTTPS connections, then Coder will proxy requests to HTTPS
+endpoints using the HTTP proxy. The previous examples will proxy all requests,
+regardless of protocol (HTTP or HTTPS), through the defined proxy.
+
+To configure a different proxy to use for outbound HTTPS connections, you may
+specify the same proxy types (`http`, `https`, `socks5`) using the
+`coderd.proxy.https` key.
```
johnstcn, line 56: Add an example?

guides/deployments/proxy.md
```diff
@@ -0,0 +1,97 @@
...
+[Reverse proxies](#reverse-proxies).
+
+## Forward proxies
+
+Coder supports proxies for outbound HTTP and HTTPS connections by configuring
+the `coderd.proxy.http` and `coderd.proxy.https` settings in the Helm chart,
+which correspond to the standard `http_proxy` and `https_proxy` environment
+variables, respectively.
+
+If the proxy URL does not include a scheme, Coder will default to treating it as
```
johnstcn, line 20: Suggestion: active voice

```suggestion
If the proxy URL does not include a scheme, Coder will treat it as a HTTP proxy by default.
```

guides/deployments/proxy.md
```diff
@@ -0,0 +1,97 @@
...
+an HTTP proxy. It is also possible to connect to the proxy using HTTPS or SOCKS5
```
johnstcn, line 21: Suggestion: active voice

```suggestion
Coder also supports proxies using the HTTPS or SOCKS5 protocols.
```

guides/deployments/proxy.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Reverse proxies
+
+If you have a reverse proxy in front of Coder, as will be the case when you are
+using an Ingress Controller, Coder will receive connections originating from the
+proxy. In order for auditing, logging, and other features to associate the
+connecting user's IP address information correctly, you will need to configure
+the `coderd.reverseProxy` setting.
+
+By default, to prevent clients from spoofing their originating IP addresses
```
johnstcn, line 77: Suggestion: split long sentence.

```suggestion
Coder will, by default, ignore `X-Forwarded-For` or similar headers and remove them from proxied connections to [Dev URL Services](../../workspaces/devurls.md). This is in order to prevent clients from spoofing their originating IP addresses.
```

johnstcn (approved):
Thanks for this! I have some suggestions, but defer to @khorne3

manifest.json
khorne3: It was an accident :) I updated the list to be alphabetical

khorne3 (approved):

khorne3:
@jawnsy sorry for the delay, I lost track of this! :(

guides/deployments/proxy.md
khorne3: @jawnsy ☝️

jawnsy: added!

manifest.json
jawnsy: Cool, thanks for confirming!

guides/deployments/proxy.md
```diff
@@ -18,27 +18,36 @@ configured the `coderd.proxy.http` and `coderd.proxy.https` settings in the
 [Helm chart](../admin/helm-charts.md). These settings correspond to the standard
 `http_proxy` and `https_proxy` environment variables, respectively.
 
-If the proxy URL does not include a scheme, Coder defaults to treating it as
-an HTTP proxy. Coder also supports proxies using HTTPS or SOCKS 5 protocols.
+If the proxy URL does not include a scheme, Coder defaults to treating it as an
+HTTP proxy. Coder also supports proxies using HTTPS or SOCKS 5 protocols. As a
+special case, Coder will always establish connections to `localhost` directly,
```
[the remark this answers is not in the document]
jawnsy: there's two protocols involved here: the protocol of the endpoint we're trying to connect to (HTTP or HTTPS), and the protocol of the proxy (HTTP, HTTPS, SOCKS 5)

what this sentence was trying to indicate is that, when trying to connect to a `localhost` endpoint - regardless of protocol (HTTP or HTTPS) or port number of the endpoint - we connect directly and do not use the proxy

the reason for this is that, we're connecting to a remote proxy and then asking it to connect somewhere else. if we connect to the proxy and ask for localhost, then it will connect us to the host running the proxy, rather than the local machine

PR #833, opened 2022-02-07
chore: fix italics to stop linter from breaking everything

khorne3:


jawnsy (approved):
weird linter error, but change LGTM

PR #842, opened 2022-02-11
feat: add GeoDNS documentation

deansheather:


deansheather:
Very rough. I'm not super happy with the Cloudflare section taking up the whole document but I'm unsure how to solve it

admin/satellites/geo-dns.md
```diff
@@ -0,0 +1,120 @@
+---
+title: GeoDNS
```
jawnsy, line 2: How about we call this "global load balancing," or something similar?

If you google GeoDNS, this is the top hit: https://en.wikipedia.org/wiki/GeoDNS

Do these docs only refer to Anycast routing, or can people use this if they are using split-horizon DNS?

admin/satellites/geo-dns.md
```diff
@@ -0,0 +1,120 @@
...
+description: |
+    Learn how to configure your primary and satellite deployments to share a
+    single hostname.
+state: alpha
+---
+
+Out of the box, satellites deployments have a different access URL to the
+primary deployment. This may cause confusion for engineering teams located
```
jawnsy, line 10: This says "This may cause confusion" twice

admin/satellites/geo-dns.md
```diff
@@ -0,0 +1,120 @@
...
+globally. This may cause confusion amongst users wondering which access URL they
+should access Coder from.
+
+Coder supports a GeoDNS configuration where all satellites and the primary
+deployment share the same hostname, so all users access Coder using a consistent
```
jawnsy, line 15: needs a passage about users being routed to the nearest installation or satellite, ensuring low latency

admin/satellites/geo-dns.md
```diff
@@ -0,0 +1,120 @@
...
+
+> If you are using cert-manager, you can easily add more hostnames to your
+> certificate by adding them to the `spec.dnsNames` section.
+
+To create a GeoDNS load balancer on Cloudflare, you can refer to our simple
+instructions [below](#create-a-geodns-load-balancer-on-cloudflare).
+
+## Configure GeoDNS on Coder
+
+To configure GeoDNS onn Coder:
```
jawnsy, line 47:
```suggestion
To configure GeoDNS on Coder:
```

admin/satellites/geo-dns.md
```diff
@@ -0,0 +1,120 @@
...
+
+1. In the primary helm values file set `coderd.alternateHostnames` to your
```
jawnsy, line 49:
```suggestion
1. In the primary helm values file, set `coderd.alternateHostnames` to your
```

admin/satellites/geo-dns.md
```diff
@@ -0,0 +1,120 @@
...
+       your GeoDNS hostname:
+
+       ```yaml
+       coderd:
+         alternateHostnames:
+           - "satellite.example.com"
+           - "coder.example.com"
+       ```
+
+1. Deploy your primary and satellites with your new helm values.
```
jawnsy, line 72:
```suggestion
1. Deploy your primary and satellites with your new Helm values.
```

admin/satellites/geo-dns.md
```diff
@@ -0,0 +1,120 @@
...
+
+   ```yaml
+   coderd:
+     alternateHostnames:
+       - "primary.example.com"
+       - "coder.example.com"
+   ```
+
+1. In each satellite helm values file:
+    1. Set `coderd.satellite.accessURL` to your GeoDNS URL (this value is used
```
jawnsy, line 60:
```suggestion
    1. Set `coderd.satellite.accessURL` to your primary access URL, which is the name of the global load balancer in front of Coder deployments (this value is used
```

admin/satellites/geo-dns.md
```diff
@@ -0,0 +1,120 @@
...
+> through this configuration process.
+
+## Requirements
+
+You will need the following:
+
+- A primary access URL (e.g. `https://primary.example.com`).
+- One or more satellite access URLs (e.g. `https://sydney.example.com`,
+  `https://london.example.com`).
+- A GeoDNS load balancer access URL (e.g. `https://coder.example.com`). The
```
jawnsy, line 30: GCP does something similar with Anycast, https://cloud.google.com/load-balancing

We should distinguish whether this feature is for people using Anycast (a CDN or global load balancing) versus split-horizon DNS (where the DNS responses vary according to the user's IP address/location)

khorne3:
@deansheather I don't think the Cloudflare section is too big of a problem, but if you'd like, we can split it into its own guide? Then people could use if if they need the help in that section, otherwise they can ignore if they have a load balancer already

khorne3:
@deansheather do you have any thoughts on @jawnsy 's suggestions?

- Calling this feature `global load balancing` or similar
- Distinguish whether this feature is for people using Anycast (a CDN or global load balancing) versus split-horizon DNS (where the DNS responses vary according to the user's IP address/location)

Other than that, I think all else is ok

deansheather:
@jawnsy GeoDNS is a correct term for it AFAIK and means the same thing as split-horizon DNS. What are your thoughts on disambiguating it in the first paragraph but keeping it called GeoDNS?

deansheather:
@khorne3

- I can rename it to "global load balancing" if you think it's a good idea, GeoDNS is a good name though IMO
- I added a paragraph that says it will work with both anycast and split-horizon

deansheather:
I've renamed it to "Proximity based configuration" but I don't think that's a good name yet. Also, I've replaced most of the references to "GeoDNS access URL" with "unified access URL" instead

jawnsy (approved):

PR #847, opened 2022-02-14
chore: add info re: breaking changes in 1.27.x

khorne3:


changelog/1.27.0.md
```diff
@@ -5,7 +5,32 @@ description: "Released on 01/19/2022"
 
 ### Breaking changes ❗
 
-There are no breaking changes in 1.27.0.
+- Users upgrading from Coder v1.21 (or earlier) *must* update their Helm values
```
deansheather, line 8: It's not for users upgrading from 1.21, it's more for users with deployments that predate 1.21. I.e. you could have gone 1.20 -> 1.21 -> 1.23 -> 1.25 -> 1.27 and you could still be affected by this if you haven't upgraded your values already. We have been warning this since 1.21 in the `helm upgrade` output

deansheather: sounds good but maybe we should include the "if you haven't already" part too

jawnsy (approved):
Thank you for this!

PR #854, opened 2022-02-16
chore: update Docker versions

khorne3:
I made a note on the release tasklist to update these, too

### DIFF
--- a/setup/docker.md
+++ b/setup/docker.md
@@ -88,7 +88,7 @@ DEVURL_HOST="*.mycompany.com"
 For example:
 
 ```console
-docker run --rm -it -p 7080:7080 -v /var/run/docker.sock:/var/run/docker.sock -v ~/.coder:/var/run/coder -e DEVURL_HOST="*.mycompany.com" codercom/coder:1.27.0
+docker run --rm -it -p 7080:7080 -v /var/run/docker.sock:/var/run/docker.sock -v ~/.coder:/var/run/coder -e DEVURL_HOST="*.mycompany.com" codercom/coder:1.28.0
 ```
 
 ## Use an external PostgreSQL database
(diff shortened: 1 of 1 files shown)


vapurrmaid (approved):

setup/docker.md
```diff
@@ -88,7 +88,7 @@ DEVURL_HOST="*.mycompany.com"
 For example:
 
 ```console
-docker run --rm -it -p 7080:7080 -v /var/run/docker.sock:/var/run/docker.sock -v ~/.coder:/var/run/coder -e DEVURL_HOST="*.mycompany.com" codercom/coder:1.27.0
+docker run --rm -it -p 7080:7080 -v /var/run/docker.sock:/var/run/docker.sock -v ~/.coder:/var/run/coder -e DEVURL_HOST="*.mycompany.com" codercom/coder:1.28.0
```
jawnsy, line 91: I think there's a point release coming soon, @deansheather @Emyrk should we use 1.28.1 here?

We may want to consider publishing a `latest` tag so that we don't have to update the docs, but I defer to Dean/Steven on that

jawnsy (approved):
LGTM!

setup/docker.md
deansheather: it'd be nice if there were a template value used in the docs to get the latest version or something

khorne3: will update if there's a point release!

jawnsy: Usually people publish multiple tags with semver:

* `codercom/coder:latest` - the default version that people should use if they don't otherwise specify, usually the most recent stable version
* `codercom/coder:1` - the most recent version of 1.x
* `codercom/1.28` - the most recent version of 1.28, which might be 1.28.0, 1.28.1, 1.28.2, etc

This lets people choose their own tradeoff in terms of automatic updates vs stability/reproducibility. For an example, see the official alpine image's tags: https://hub.docker.com/_/alpine?tab=tags

PR #865, opened 2022-02-25
chore: create CODEOWNERS

vapurrmaid:


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+@coder/coder-docs
```
jawnsy, line 1: does this ping coder-docs for all changes in this repo?

jawnsy (approved):
LGTM, just had a minor question

.github/CODEOWNERS
vapurrmaid: Yea, which might not be the best granularity (we may want to separate owners for CI or other aspects). Very open to ideas, I opened this to start that convo ✨

khorne3 (approved):

.github/CODEOWNERS
khorne3: I think it should be fine? It's just us at this moment and I can handle the (infrequent CI) stuff. It's not the busiest of repos

PR #869, opened 2022-03-01
chore: add workspace setup guidance

khorne3:


guides/workspaces.md
```diff
@@ -0,0 +1,87 @@
+---
```
jawnsy, line 1: This is missing an entry in the manifest, so it's currently hidden from the navbar. Is there some other way I can preview this? Going here just returns a 404: https://codercom-qpu8omkyu-codercom.vercel.app/docs/coder/v1.28/guides/workspaces

guides/workspaces.md
```diff
@@ -0,0 +1,87 @@
...
+one workspace for your Python projects, and so on.
+
+Upsides:
+
+- Streamlined images due to them only containing only one language and its
+  accompanying dependencies
+
+Downsides:
+
+- Developers potentially have multiple workspaces, each of which are are the
```
jawnsy, line 43: "Developers may have multiple workspaces, consuming more storage space overall" ?

jawnsy (approved):
Content generally LGTM, thanks for writing this! I had some minor suggestions for you to consider, mainly to try to simplify the language and avoid jargon

guides/workspaces.md
```diff
@@ -0,0 +1,85 @@
+---
+title: "Workspace organization"
+description: Learn how to organize your Coder workspaces.
+---
+
+This article describes considerations for deciding how to set up your Coder
+workspaces.
+
+In general, the fewer workspaces per developer, the easier it is for the
+individual developer to manage. However, the complexity of the underlying images
```
jawnsy, line 10: Maybe add a sentence/link like: "For teams that do not have a complex development workflow, we recommend [one workspace per developer], as it is easiest to adopt."

guides/workspaces.md
```diff
@@ -0,0 +1,85 @@
...
+you would a laptop: the workspace is where you have all of your languages,
+dependencies, and tooling installed, and it is the one place you'd go to work on
+your projects.
+
+Benefits:
+
+- Fewer workspaces to manage
+- No need to switch between workspaces for different projects
+
+Limitations:
```
jawnsy, line 26: Actually thinking about this more, maybe instead of Limitations, "Caveats" could be better? These aren't always disadvantages, since they only apply particular cases (hence "can" vs "will")... ideally, users that have very simple workflows will enjoy the benefits and won't be affected by the potential limitations. Maybe "Potential caveats" even?

PR #881, opened 2022-03-08
feat: coderd resource recommendations

ericpaulsen:
adding context around `coderd` reqs/limits and replicas.

### DIFF
--- a/setup/requirements.md
+++ b/setup/requirements.md
@@ -3,11 +3,33 @@ title: "Requirements"
 description: Learn about the prerequisite infrastructure requirements.
 ---
 
+### Compute
+
 Coder is deployed onto Kubernetes clusters, and we recommend the following
 resource allocation minimums to ensure quality performance.
 
-For **basic control services**, allocate at least 2 CPU cores, 4 GB of RAM, and
+For the Coder control plane (which consists of the `coderd` pod and any
+additional replicas) allocate at least 2 CPU cores, 4 GB of RAM, and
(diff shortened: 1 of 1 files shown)


setup/requirements.md
```diff
@@ -3,12 +3,34 @@ title: "Requirements"
...
+resources:
+    requests:
+      cpu: "250m"
+      memory: "512Mi"
+    limits:
+      cpu: "250m"
+      memory: "512Mi"
+```
+
+The default number of `coderd` replicas is 1.
```
jawnsy, line 29:
```suggestion
By default, Coder is a single-replica deployment. For production systems, consider using at least 3 replicas to provide failover and load balancing capabilities.
```

Generally, high availability topologies require at least 3 replicas (with pod affinity settings configured to spread the pods across nodes, which is the default). This makes the deployment tolerant to node failure as well as application crashes. Three replicas is a good minimum because, if a single replica in a three-replica configuration fails, the remaining two replicas will receive a 50% increase in load, rather than a doubling (as would occur with a two-replica configuration)

setup/requirements.md
```diff
@@ -3,12 +3,34 @@ title: "Requirements"
...
+
+If you expect roughly 10 or more concurrent users, we recommend increasing these
```
jawnsy, line 31: Maybe also worth documenting that we recommend people do performance testing in a staging environment

khorne3 (approved):

PR #896, opened 2022-03-16
fix(tls): cert-manager installation script

vapurrmaid:
## Summary

We had an outdated installation script for cert-manager

## See also

https://cert-manager.io/docs/installation/

## Note

We don't have any checks/process in place for checking on this script (ex: parsing it from https://cert-manager.io/docs/installation/). We are prone to getting out of date again without some kind of process around it.


vapurrmaid:
[sc-22225]

PR #900, opened 2022-03-16
fix: link to version support policy

vapurrmaid:
[sc-21348]


vapurrmaid:
Hrm, I wonder why this is not deploying for preview?

khorne3:
Looks like either Vercel or GH Actions needed more coffee this AM, but better late than never?

khorne3 (approved):

vapurrmaid:
The link (version support policy) at the top of 1.28 works in the preview: https://coder.com/docs/coder/9c2b1d4/changelog/1.28.0

and this seems to fix it in-product

PR #909, opened 2022-03-17
chore: add 1.28.4 changelog

khorne3:


changelog/1.28.4.md
```diff
@@ -0,0 +1,21 @@
...
+description: "Released on 03/17/2022"
+---
+
+### Breaking changes ❗
+
+There are no breaking changes in 1.28.4.
+
+### Features ✨
+
+- web: added ability to host the **Open in Coder** SVG image in Coder in
```
khorne3, line 12: @vapurrmaid wdyt?

vapurrmaid: We now use the SVG hosted in Coder by default, so we might want to capture a few things here:

- a) this is a fix for firewalled environments that can't access an external cdn
- b) this is the new default, but our CDN will continue to be supported with no deprecation timeline as of yet.

I'm not sure how to word it, but here's a rough idea:


```
web: the **Open in Coder** SVGs are now statically hosted in Coder and will render properly in firewalled networks. Our external CDN will continue to be supported, but we default to pulling the SVG from Coder.
```

(feat --> fix)

khorne3: Not sure if the first use of static is necessary, but what about

```
web: added [static] hosting for the **Open in Coder** SVGs in Coder so that the icons render properly in firewalled networks. Coder will continue to support the existing external CDN for such icons, but the default will be the SVG hosted by Coder.
```

vapurrmaid: I think either/or works, and since brevity is helpful in a changelog, we can do it without `static`. This is great!!!

Always love your way with words @khorne3 :taco: :tada: !

vapurrmaid (approved):
✅ with the changes in https://github.com/coder/docs/pull/909/files#r829544797!

changelog/1.28.4.md
```diff
@@ -0,0 +1,24 @@
...
+### Breaking changes ❗
+
+There are no breaking changes in 1.28.4.
+
+### Features ✨
+
+- web: added hosting for the **Open in Coder SVGs** in Coder so that the
+  icons render properly in firewalled networks. Coder will continue to support
+  the existing external CDN for such icons, but the default will be the SVG
+  hosted by Coder.
```
deansheather, line 15: The wording for this seems confusing, how about:

> web: the **Open in Coder** SVG is now served from your Coder deployment rather than our CDN so airgapped customers can use this feature. Existing button images will continue to be served from our CDN until updated.

deansheather (approved):

PR #912, opened 2022-03-19
feat: Config steps for multiple JetBrains IDEs in a browser

mtm20176:
FEAT: new doc showing how to configure a custom image with a configure script and a Workspace Application config.yaml to show multiple JetBrains IDEs in a workspace so multiple projects can be opened.


ericpaulsen (approved):
Looks good to me! I will leave the word-smithing for @khorne3

PR #950, opened 2022-04-18
add: default user session expiration

ericpaulsen:
adds info on the default user session expiration.

### DIFF
--- a/admin/access-control/index.md
+++ b/admin/access-control/index.md
@@ -14,10 +14,16 @@ to manually create users who log in with their email address and temporary
 password. Coder will ask them to change their password after they log in the
 first time.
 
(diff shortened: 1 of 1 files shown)


admin/access-control/index.md
```diff
@@ -14,11 +14,17 @@ to manually create users who log in with their email address and temporary
 password. Coder will ask them to change their password after they log in the
 first time.
 
+The default user session expiry time is 1 week for users logging in via built-in
```
khorne3, line 17:
```suggestion
The default user session expiry time is one week for users logging in via built-in
```

khorne3 (approved):

mtm20176 (approved):
Good addition. Comes from customer feedback who asked the question.

PR #1001, opened 2022-05-27
chore: remove mentions of CNIs, Calico

khorne3:


ericpaulsen:
@khorne3 - i think we should close this PR. i've confirmed that a CNI that enforces network policies is required for the Coder-created network policies to work properly.

Now, that doesn't necessarily mean that installing Calico is _required_, but it is recommended, given that it supports `NetworkPolicy` enforcement (and the AWS native VPC CNI does not). I'm thinking we downgrade the Calico "step" to a recommendation in the AWS docs.

Notably, the [native Azure CNI supports](https://docs.microsoft.com/en-us/azure/aks/use-network-policies#network-policy-options-in-aks) `NetworkPolicy` enforcement, so (in a separate PR) we can change the flag to `--network-policy azure`.

For Google, the `--enable-network-policy` flag enables Calico by default, so we should keep this in place.

PR #1029, opened 2022-06-22
chore: Change openshift example to use userdel/useradd instead of usermod

holzman:
When using usermod (or useradd without -l), a sparse file is created
for lastlog/faillog.  The extents of the file scale with the UID, and
Docker does not see the file as sparse, so it will try to generate a
huge image without this patch.

(See https://github.com/moby/moby/issues/5419 for details)

Signed-off-by: Burt Holzman <burt@fnal.gov>


Emyrk (approved):
Good find :+1:

PR #1037, opened 2022-07-09
chore: Setup docs for Kubernetes on Google Cloud - to handle the cloud provider's required node image change from `Docker` to `containerd`

mark-theshark:
A prospect was blocked by the existing GKE install scripts which use Docker-based node images. e.g., ubuntu and cos.

The latest versions of Kubernetes require `containerd`-based node images, so the `image-type` must change to either `ubuntu_containerd` or `cos_containerd`

I also consolidated the machine-type overview, and changed to a more economical e2-standard-4 with nodes with links to general-purpose machine-type docs on Google Cloud.

Reference article (which I linked to in the docs) about the `containerd` change:
https://cloud.google.com/kubernetes-engine/docs/concepts/node-images

### DIFF
--- a/setup/kubernetes/google.md
+++ b/setup/kubernetes/google.md
@@ -24,6 +24,10 @@ be sure to account for the number of developers you expect to use Coder, as well
 as the resources they need to run their workspaces. See our guide on on
 [compute resources](../../guides/admin/resources.md) for additional information.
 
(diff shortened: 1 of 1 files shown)


ericpaulsen (approved):
Looks good. I believe this is a result of Dockershim deprecation in k8s 1.24 - https://kubernetes.io/blog/2022/02/17/dockershim-faq/

PR #1074, opened 2022-08-09
add: JetBrains IDE/Projector incompatabilities

ericpaulsen:
makes known an issue with JetBrains IDEs version `2022.2` being incompatible with the Projector client/server versions used in Coder releases `1.33.0` & earlier.


ericpaulsen:
@bpmct if we're planning on back-porting the Projector client/server updates, then we won't need to mark this as a known issue. thoughts?

bpmct:
Makes sense to me, not sure when the back-port is going to come out but I bumped it.

bpmct (approved):

PR #1136, opened 2022-09-30
Updating AWS RDS IAM Auth Documentation

phelewski:
I ran into an issue recently when trying to enable this for a new RDS DB Cluster within AWS. My proposed changes are what got me to a working solution with my setup. This included updating the Trust Relationship policy condition to only use the IAM Identity Provider name instead of the full ARN. Also, I ran into a permission issue with a fresh database and my non-admin database user, so I needed to provide an extra grant command to allow the ability to create expressions as part of the initial startup.


ericpaulsen (approved):
thank you for submitting this fix!

PR #1146, opened 2022-10-07
add: admin field/audit log mappings

ericpaulsen:
adds tables for the expected outputs of admin setting changes in the audit log.

### DIFF
--- a/admin/audit.md
+++ b/admin/audit.md
@@ -63,7 +63,9 @@ actions included:
 
 With the exception of a few, logged events made by Admin panel changes will output
 the changed field(s) and the new, corresponding value. Below is the expected
-(example) output for each Admin panel change:
+(example) output for each Admin panel change.
+
+> The Admin fields not documented below currently do not output a field/diff.
 
 ### Infrastructure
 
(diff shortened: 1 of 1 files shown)


admin/audit.md
```diff
@@ -58,3 +58,75 @@ actions included:
 - `view`: the Coder CLI used a secret
 - `write`: the user made a change to a Coder entity (e.g., workspace, user,
   resource pool, etc.)
+
+## Admin logged events
+
+With the exception of a few, logged events made by Admin panel changes will output
```
bpmct, line 64: Should we document the exception?

bpmct (approved):
This is super thorough! Awesome addition to the docs.

Do we have a plan for how we update this when new features are added? It may be infrequent, but perhaps code comments or something would help us remember to update this?

admin/audit.md
ericpaulsen: we should, but those will be added soon, so we'd have to go back and remove it. i'm thinking whatever is not documenting isn't logged. what do you think?

bpmct: that's fair.

PR #1150, opened 2022-10-13
minor fix: images icon alignment

antonkor:
sidebar icon alignment fix


antonkor:
nothing major here, just found this icon not aligned vertically: 
image

this will fix it look like this: 
image

bpmct (approved):

PR #1177, opened 2022-11-17
feat: Add docs about workspace metric cardinality

Emyrk:
Future work can be to add the workspace metrics like coderd metrics

### DIFF
--- a/admin/prometheus.md
+++ b/admin/prometheus.md
@@ -32,7 +32,63 @@ spec:
     - port: prom-coderd
 ```
 
-## Metrics
+## Workspace Metrics
+
+Each coder workspace has an agent that connects to a single `coderd` instance.
+Each coderd instance will include all metrics from the workspaces it manages.
+The workspace metrics will all look like this:
+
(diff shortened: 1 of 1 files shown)


admin/prometheus.md
```diff
@@ -32,7 +32,63 @@ spec:
...
+coder update, etc), the metrics for that workspace will be moved to the new
+`coderd` metrics endpoint. The labels on the new metrics will likely have the
+new `coderd` pod name. So when tracking a singular workspace, you should track
+**only** by `workspace_id` throughout the lifetime of the workspace until it is
+deleted.
+
+### Drop workspace metrics config
+
+[Prometheus Documentation](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs)
+about relabling metrics. In this case we will drop all metrics that contain the
```
johnstcn, line 61:
```suggestion
about relabelling metrics. In this case we will drop all metrics that contain the
```

johnstcn (approved):

PR #1181, opened 2022-11-26
fix: manifest.json: add v1.37 to versions array

johnstcn:
Add missing v1.37 to versions array.
Also add a script `validate-manifest` to ensure that missing versions in the changelog are caught by `lint-staged`.
TODO: cherry to v1.37 branch


ericpaulsen (approved):

ericpaulsen:
@johnstcn it's showing the commit hash instead of the version:

Screen Shot 2022-11-27 at 7 17 55 PM

johnstcn:
> @johnstcn it's showing the commit hash instead of the version:
> 
> Screen Shot 2022-11-27 at 7 17 55 PM

I believe that's normal on preview.

scripts/validate-manifest.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+#
+# validate-manifest.sh [path/to/manifest.json]
+#
+# Dependencies: jq tr sort uniq
+#
+# Description: Ensures consistency of versions specified in manifest.json.
+
+set -eou pipefail
```
mafredri, line 9: Pipefail is a parameter to `-o`, if the previous worked I can imagine it might fail in some versions of bash.

```suggestion
set -euo pipefail
```

scripts/validate-manifest.sh
```diff
@@ -0,0 +1,44 @@
...
+
+if ! command -v jq > /dev/null; then
+    echo "This script requires jq to be available."
+    exit 1
+fi
+
+GIT_ROOT=$(git rev-parse --show-toplevel)
```
mafredri, line 16: This script could be run outside project root (at least theoretically, getting the script dir and cd-ing there could be useful).

scripts/validate-manifest.sh
```diff
@@ -0,0 +1,44 @@
...
+MANIFEST_JSON_PATH=${1:-"${GIT_ROOT}/manifest.json"}
+
+declare -a MANIFEST_VERSIONS
+declare -a ROUTE_VERSIONS
+
+# Read the versions array in manifest.json
+readarray -t MANIFEST_VERSIONS < <(jq -r '
```
mafredri, line 23: I believe `readarray` sets a bash version requirement, but not sure when it was added.

scripts/validate-manifest.sh
```diff
@@ -0,0 +1,44 @@
...
+    .routes[]
+    | select(.path == "./changelog/index.md")
+    | .children[]
+    | .path |
+    capture("(?<v>[0-9]+\\.[0-9]+)")
+    | .v' < "${MANIFEST_JSON_PATH}")
+
+# Compare the two
+DIFF=$(echo "${MANIFEST_VERSIONS[@]}" "${ROUTE_VERSIONS[@]}" | tr ' ' '\n' | sort | uniq -u)
+if [[ -n $DIFF ]]; then
```
mafredri, line 39: This seems to be missing an actual `diff` command? I imagine `$DIFF` will always be non-empty.

johnstcn: The `uniq -u` does the diff:

```shell
++ echo 1.37 1.36 1.35 1.34 1.33 1.32 1.31 1.30 1.37 1.36 1.35 1.34 1.33 1.32 1.31 1.30
++ tr ' ' '\n'
++ sort
++ uniq -u
+ DIFF=
+ [[ -n '' ]]
+ exit 0
```

mafredri: Ohh, TIL. Nice!

scripts/validate-manifest.sh
johnstcn: Seems to have been added in 4.x; added a preflight check for Bash version.

mafredri (approved):

PR #1200, opened 2023-01-03
feat: Add the new organization roles to the docs

Emyrk:


johnstcn (commented):
Also need to update the table starting on line 9, as well as the text above it "There are two roles available:"

Emyrk:
Just did that :+1:

johnstcn (approved):

PR #1252, opened 2023-04-07
Correct rsync command

embold-tyler:
The previous command example with `user@coder.<workspace>` does not work and gives a `fatal: failed to find workspace` error.


johnstcn (approved):

johnstcn:
Thanks @embold-tyler !

PR #1320, opened 2023-10-04
Remove mention of legacy extension marketplace

code-asher:
We would like to get this shut down.

Also the environment variable changed a long while back.


code-asher:
I have this against main but I am not sure that is the right branch, I see some 1.44 stuff has not merged into main yet so maybe this PR should use that as the base?

